PR #24, opened 2018-02-02
Tighten up deploy a bit

jpstrikesback:
This adds the `serviceRoleArn` var to get-config's returns,
grabs the missing `userPoolConfig`,
swaps in the `definition` key and gives it Buffer.from(the schema) in createGraphQLSchema(),
and swaps `async.whilst` for `async.until` in monitorGraphQLSchemaCreation()

### DIFF
--- a/serverless-appsync-plugin/get-config.js
+++ b/serverless-appsync-plugin/get-config.js
@@ -50,6 +50,7 @@ module.exports = (config, provider, servicePath) => {
     authenticationType: config.authenticationType,
     schema: schemaContent,
     userPoolConfig: config.userPoolConfig,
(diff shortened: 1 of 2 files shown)


sid88in:
awesome BIG thanks @jpstrikesback for making the first contribution. did you try installing this plugin? We are still polishing it but would <3 even more contributions, so that we can officially release it :)

jpstrikesback:
Cheers, and thanks for making this! Yeah, installed and playing with it, I’ll be playing with the resolver mapping’s in the plugin next, to try to generalize that a bit.

sid88in:
@jpstrikesback that's great! thanks for being our first user :D

PR #41, opened 2018-02-10
Update README.md

jpstrikesback:
Fix examples in readme 😀


sid88in:
I did the same ;) not required

PR #47, opened 2018-02-14
v0.0.4 - Added cleanupDataSources function

lolcoolkat:
- Added cleanupDataSources that removes any unused data sources when updating. For example if you have a data source called `Lambda` on your API and it isn't defined in your serverless.yml file .. it will be deleted. 

Updated to 0.0.4


sid88in:
great 🎉

PR #51, opened 2018-02-17
add api key functionality

sid88in:


sid88in:
need to add the following condition:

IF resolvedConfig.authenticationType == 'API_KEY'
THEN createAPIKey()

Also, createAPIKey() needs to be called after createGraphQLEndpoint() to fetch the api id.

sid88in:
@lolcoolkat lets discuss this design - few options:

1. create the API key in createGraphQLEndpoint itself if authenticationType == API_KEY. In this case it probably will have to be done after promise is resolved because we would need api id

2. keep creation out the api key outside createGraphQLEndpoint and make it one of the steps and bind it in the hook (one issue here is that we would not want to create api key in case authenticationType != API_KEY. 

Any thoughts?

sid88in:
thinking more on this - use should be able to pre-configure API KEY and in that case createKey might not be required!

RafalWilinski:
I think that:
```js
if (resolvedConfig.authenticationType === 'API KEY') {
  this.createAPIKey();
}
```

inside `createGraphQLEndpoint` `then` is totally fine. 

Moreover, `userPoolConfig` param should look something like this:
```js
userPoolConfig: resolvedConfig.authenticationType === 'AMAZON_COGNITO_USER_POOLS' && {
  awsRegion: resolvedConfig.region,
  defaultAction: resolvedConfig.userPoolConfig.defaultAction,
  userPoolId: resolvedConfig.userPoolConfig.userPoolId
}
```

sid88in:
AWS just launched new functionalities for API KEY https://aws.amazon.com/blogs/mobile/aws-appsync-region-expansion-and-new-features-for-api-keys/

@RafalWilinski you are correct. Also, looks like we can use updateApiKey to attach existing API Key
because if I already have an API KEY, then I don't want to use createAPIKey yeah?

PR #54, opened 2018-03-07
follow eslint rules

jpstrikesback:
This change:

- formats code to follow eslint
- fixes two catch errors where the log was using a nonexistent variable

I would be very happy to see prettier used instead :) :)


sid88in:
@jpstrikesback totally.

PR #85, opened 2018-04-29
generate mapping templates for fields

merunga:
Given a `fields` property in the the data source conf, autogenerate the mapping templates. For example

```
dataSources:
   - type: AWS_LAMBDA
     name: Lambda
     description: 'Lambda DataSource'
     config:
       lambdaFunctionArn: ...
       serviceRoleArn: ...
     fields:
       - listEvents
       - commentOnEvent
```

It works only for lambda data sources for now

### DIFF
--- a/lib/codegen/AWS_LAMBDA-request-mapping-template.tmpl.ejs
+++ b/lib/codegen/AWS_LAMBDA-request-mapping-template.tmpl.ejs
@@ -1,0 +1,8 @@
+{
+  "version": "2017-02-28",
+  "operation": "Invoke",
+  "payload": {
+      "field": "<%= field %>",
+      "arguments":  $utils.toJson($context.arguments)
+  }
+}
--- a/lib/codegen/AWS_LAMBDA-response-mapping-template.tmpl.ejs
+++ b/lib/codegen/AWS_LAMBDA-response-mapping-template.tmpl.ejs
@@ -1,0 +1,1 @@
+$util.toJson($context.result)
--- a/lib/codegen/index.js
+++ b/lib/codegen/index.js
@@ -1,0 +1,62 @@
+const fs = require('fs');
+const { join } = require('path');
+
+const { introspectSchema } = require('apollo-codegen');
+const ejs = require('ejs');
+
+const generateVelocityTemplate = ({
+  type = 'request', dataSourceType = 'AWS_LAMBDA', field, mappingTemplatesLocation,
+}) => {
+  const template = fs.readFileSync(join(__dirname, `${dataSourceType}-${type}-mapping-template.tmpl.ejs`), 'utf-8');
+  const mtContents = ejs.render(template, field ? { field } : {});
+  const outPath = join(mappingTemplatesLocation, `${field ? `${field}-` : ''}${type}-mapping-template.txt`);
+  fs.writeFileSync(outPath, mtContents, 'utf-8');
+};
+
+const getFieldType = async (schemaLocation, field) => {
+  const appSyncDir = join(__dirname, '.appsync');
+  if (!fs.existsSync(appSyncDir)) {
+    fs.mkdirSync(appSyncDir);
+  }
+
+  await introspectSchema(
+    schemaLocation,
+    join(appSyncDir, 'schema.json'),
+  );
+
+  /* eslint-disable no-underscore-dangle */
+  /* eslint-disable global-require */
+  const schema = require('./.appsync/schema.json').data.__schema;
+  /* eslint-enable global-require */
+  /* eslint-enable no-underscore-dangle */
+
+  let type;
+  /* eslint-disable array-callback-return */
+  schema.types.map((y) => {
+    if (
+      !type &&
+      (y.name === schema.queryType.name ||
+      y.name === schema.mutationType.name)
+    ) {
+      // console.log(y);
+      const found = y.fields.find(f => f.name === field);
+      if (found) {
+        /* eslint-disable no-nested-ternary */
+        type = y.name === schema.queryType.name ? 'Query' : (
+          y.name === schema.mutationType.name ? 'Mutation' : 'Subscription'
+        );
+        /* eslint-enable no-nested-ternary */
+      }
+    }
+  });
+  /* eslint-enable array-callback-return */
+  if (!type) {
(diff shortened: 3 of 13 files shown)


sid88in:
@merunga can you please add some screenshots to show E2E testing on AppSync console with example.

ajmath:
@merunga Why not parse the fields list from the schema, similar to what you did in this gist? https://gist.github.com/merunga/2b6d2292c5e04417c1649106071bbdcd

honkskillet:
Also, it should be ok if it is breaking

merunga:
@ajmath we talked with @sid88in about doing it like that, but that approach would only work if you had only one dataset, otherwise the script wouldn't know which dataset is used to resolved which field.
The other idea was to annotate the fields, but we wanted to keep the schema definition as standard as possible

honkskillet:
@merunga by dataset do you mean datasource (using appsync's terminology). Why wouldn't you be able to specify more than one? I think the strength of this approach is that if you need to create resources that are more complex than what is supported by this plugin you can always fall back to explicitly defining 'resources:' in the yml. And everything would still end up in the same cloud formation stack.

merunga:
yes, sorry... I did mean **datasource**. I'm editing the comment right away, thanks.

You can keep doing what you were able to do before: defining datasources, your own mappings, etc.

I'll try to explain what I meant with an example

Having these 2 datasources defined in `serverless.yml`:
```yaml
custom:
  appsync:
    dataSources:
      - type: AWS_LAMBDA
        name: Lambda1
        description: 'Lambda DataSource 1'
      - type: AWS_LAMBDA
        name: Lambda2
        description: 'Lambda DataSource 2'
```
and this `schema`:
```graphql
type Event {
  # ...
}
type Query {
  getEvent(_id: ID!): Event
}
schema {
  query: Query
}
```

how could I infer if the user meant this
```yaml
custom:
  appsync:
    mappingTemplates:
      - dataSource: Lambda1
        field: getEvent
        type: Query
        request: getEvent-request-mapping-template.txt
        response: response-mapping-template.txt
```
or this
```yaml
    mappingTemplates:
      - dataSource: Lambda2
        #...
```
?

If you have only one datasource (like in the [gist](https://gist.github.com/merunga/2b6d2292c5e04417c1649106071bbdcd)), this isn't an issue at all.

ajmath:
That makes complete sense.  Thanks for clarifying

ajmath:
This might be better as a follow-on feature, but what if the there was a smart default that included all of the fields if there was only one data source defined?

honkskillet:
I don't really get the objection. That is what custom.appsync.mappingtemplates is for, right?  You can't infer from that example  regardless of whether you translate directly into cloudformation yml or use another method.

honkskillet:
Again, if this plugin went the cloudformation route, the plugin could create the most common use cases. In more complex use cases user's could augment the plugin via directly adding resources. It actually frees the maintainers of this repo from feeling like they have to support every completely/ edge case.

bboure:
@merunga what is the status on this?
I kinda like this approach.

Lambdas usually have the same input template. In my project I actually have only 2 templates (one for Invoke, and one for BatchInvoke) that Use for all lambdas.

What about the `BatchInvoke` use case? You PR does not seem to take that into consideration. 

Any idea on how to auto-generate dynamodb templates? That sounds more complicated because each use case is different.

lib/codegen/AWS_LAMBDA-request-mapping-template.tmpl.ejs
```diff
@@ -0,0 +1,8 @@
+{
+  "version": "2017-02-28",
+  "operation": "Invoke",
```
bboure: @merunga What about "BatchInvoke" ?

lib/codegen/AWS_LAMBDA-request-mapping-template.tmpl.ejs
```diff
@@ -0,0 +1,8 @@
...
+  "payload": {
+      "field": "<%= field %>",
+      "arguments":  $utils.toJson($context.arguments)
```
bboure: @merunga What about `source` and `ìdentity` here? 
I would simply do this: 

"payload": {
	"field": "<%= field %>",
    "context":  $utils.toJson($context)
}

You don't know what the future is made of. AWS might add new fields in `context` in the future.

lib/codegen/AWS_LAMBDA-response-mapping-template.tmpl.ejs
```diff
@@ -0,0 +1 @@
+$util.toJson($context.result)
```
bboure: My response template usually looks like this

````
#if( $context.result && $context.result.errorMessage )
    $utils.error($context.result.errorMessage, $context.result.errorType, $context.result.data, $context.result.errors)
#else
    $utils.toJson($context.result.data)
#end
````

That covers the lambda returning "controlled" errors (e.g.: validation)

For example, on a createUser mutation:

success:
````js
callback(null, {
	data: {
		firstName: "John",
		lastName: "Doe",
	}
});
````

Input error:
````js
callback(null, {
	data: {},
	errorMessage: "Some Error message",
	errorType: "ValidationError",
	errors: [
		"firstName is required",
		"lastName is required",
	]
});
````

File: lib/codegen/index.js [hunk 4]
```diff
@@ -0,0 +1,62 @@
+const fs = require('fs');
+const { join } = require('path');
+
+const { introspectSchema } = require('apollo-codegen');
+const ejs = require('ejs');
+
+const generateVelocityTemplate = ({
+  type = 'request', dataSourceType = 'AWS_LAMBDA', field, mappingTemplatesLocation,
+}) => {
+  const template = fs.readFileSync(join(__dirname, `${dataSourceType}-${type}-mapping-template.tmpl.ejs`), 'utf-8');
```
bboure: @merunga What about the possibility to specify a custom template in the yml?

bboure:
@sid88in  can you point out the Amplify  doc for this ?
Thanks

sid88in:
@bboure I believe its the codegen part here https://github.com/aws-amplify/amplify-cli 
AWS is working on adding this codegen integration in the plugin https://github.com/sid88in/serverless-appsync-plugin/issues/151

PR #87, opened 2018-04-30
Support Cognito pools and data sources in a different region to AppSync

ng-marcus:
I need to deploy to eu-west-1 for AppSync, but by resources are in a different region (eu-west-2).

Whilst the documentation suggests you can (and should) provide a region for these items, the default top level region is used instead.

This PR requires a region be specified for each data source or Cognito pool which is then used when configuring AppSync.

### DIFF
--- a/README.md
+++ b/README.md
@@ -86,6 +86,7 @@ custom:
         description: # DynamoDB Table Description
         config:
           tableName: # DynamoDB Table Name
+          awsRegion: # required # region
           serviceRoleArn: "arn:aws:iam::${self:custom.accountId}:role/dynamo-${self:custom.appSync.serviceRole}"
       - type: AMAZON_ELASTICSEARCH
         name: # data source name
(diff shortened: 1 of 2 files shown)


agonbina:
@sid88in can you please review and merge this? It is a common use case since AppSync is not available in all regions.

sid88in:
@agonbina been caught up in lot of things. I'll try best to test this next week and merge it.

ajmath:
@ng-marcus What do you think about falling back to the top-level region if the data source specific region isn't defined?

Like this:
```javascript
    if (resolvedConfig.authenticationType === 'AMAZON_COGNITO_USER_POOLS') {
       config.userPoolConfig = {
         awsRegion: resolvedConfig.userPoolConfig.awsRegion || resolvedConfig.region,
         defaultAction: resolvedConfig.userPoolConfig.defaultAction,
         userPoolId: resolvedConfig.userPoolConfig.userPoolId,
       };
    }
```

ng-marcus:
Falling back to the top level region is a good idea.  I'll add that.

sid88in:
@ng-marcus we just restructured the repository. can you please sync with master and test again?

thanks
Sid

ng-marcus:
Sorry, I was on vacation, I'll take a look.

sid88in:
sounds good @ng-marcus thanks

sid88in:
@ng-marcus any updates on this?

sid88in:
@ng-marcus closing this PR. Please create a new one when you are ready. thanks

PR #93, opened 2018-06-01
Support openid

francisu:


francisu:
Whoops, this is not in the right place. Sorry.

PR #102, opened 2018-06-11
Add support for resolvers with same field name within different types

### LINKED ISSUE #97: Unable to define two resolver instances for same name of field with different types
Following code causes that only one type would be created within CloudFormation eg. when you have for instance `Employee:company` and `Query:company` fields within your GraphQL schema only the latter would be created.

https://github.com/sid88in/serverless-appsync-plugin/blob/4273e9d142d426d4be89eecedde6f491dfca0710/index.js#L185

I was testing to change it to the following:

```js
        [`GraphQlResolver${this.getCfnName(tpl.type)}${this.getCfnName(tpl.field)}`]: {
```

so both resolver configurations would be created but the CloudFormation fired error. The error is documented within [AWS forum thread](https://forums.aws.amazon.com/thread.jspa?threadID=283583) so once it gets resolved I will introduce the pull request for this change.

ivanbarlog:
fixes #97

### DIFF
--- a/index.js
+++ b/index.js
@@ -182,7 +182,7 @@ class ServerlessAppsyncPlugin {
       const reqTemplPath = path.join(config.mappingTemplatesLocation, tpl.request);
       const respTemplPath = path.join(config.mappingTemplatesLocation, tpl.response);
       return Object.assign({}, acc, {
-        [`GraphQlResolver${this.getCfnName(tpl.field)}`]: {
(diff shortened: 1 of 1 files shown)


jdeanwaite:
An unfortunate but necessary side effect of this fix is that you must delete all existing resolvers since this will now generate new resolvers with all new names.

PR #127, opened 2018-08-03
Make Request & Response Mapping Template extension from .txt to .vtl

### LINKED ISSUE #123: Make Request & Response Mapping Template extension from .txt to .vtl
`.txt` is for text files but it makes sense to make the `example` in the project change the extension to `.vtl` because it is using Velocity Template Language.

Benefits might be if someone in future makes a extension for Syntax Highlighting, it would be much simpler to detect all `.vtl` files & apply the extension on `.vtl` file rather than a `.txt` file.

This change does not have a significant impact but it makes sense like it does for GraphQL files (`.gql` extension). What do ya'll think❓

deadcoder0904:
Closes #123


deadcoder0904:
First to commits are usual third one was a failing test Idk how it worked before if it did plz check if correct

PR #130, opened 2018-08-04
Fix broken schema.graphql path from previous merged PR

Znergy:
The previous merged PR #127, also made an addition to the `get-config.js` file which added a local path to the example project schema. Which was breaking deployments as the path to my schema.graphql file was pointing at `../serverless-appsync-plugin/example/`.

**Code Snippet causing issues:**
```
const schemaPath = path.join(
    servicePath,
    '../serverless-appsync-plugin/example/',
    config.schema || 'schema.graphql',
  );
```

**Needs to be:**
```
const schemaPath = path.join(
    servicePath,
    config.schema || 'schema.graphql',
  );
```
By deleting this addition, the deployments work again. You can test this break, by updating to the most recent version of the `serverless-appsync-plugin`.

**Break in deployment:**
screen shot 2018-08-03 at 5 26 28 pm

**Change made in #127:**
screen shot 2018-08-03 at 5 25 46 pm

### DIFF
--- a/get-config.js
+++ b/get-config.js
@@ -41,7 +41,6 @@ module.exports = (config, provider, servicePath) => {
 
   const schemaPath = path.join(
     servicePath,
-    '../serverless-appsync-plugin/example/',
     config.schema || 'schema.graphql',
   );
 


sid88in:
Thanks !!!

Znergy:
@deadcoder0904 I'm not sure why tests were failing. Maybe they are testing the wrong thing. I merely took it from the perspective of actually trying to make a deployment. When I would run `sls deploy` the `serverless-appsync-plugin` would break due to the path being a non existent file.

I would imagine that any changes to tests should not be added into `get-config.js`, but into `get-config.test.js`.

deadcoder0904:
Yeah I still think the test will fail now in the repo so that's why I changed it in the 1st place. Anyways I'm yet to try out this plugin. Thanks for sending a quick fix.

PR #140, opened 2018-08-12
Fix error when `substitutions` is omitted + fix Unit tests

bboure:
When the `substitutions` config is omitted, an error is thrown:

````
Cannot convert undefined or null to object
````

Also fixes Unit tests!

Also, May I suggest that we implement some kind of CI on this repo? e.g: circleci or travis ?

### DIFF
--- a/__snapshots__/get-config.test.js.snap
+++ b/__snapshots__/get-config.test.js.snap
@@ -118,6 +118,7 @@ schema {
 }
 ",
   "serviceRoleArn": "1234",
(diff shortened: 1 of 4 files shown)


sid88in:
@bboure I totally agree. We should have travis. There is an open task for it but just never got prioritized. Want to take that up?

bboure:
@sid88in I'm not familiar with Travis But I'll have a look 👍

Issue #154, opened 2018-09-14
Feature Request/Discussion: Support for multiple APIs


ZY-Ang:
Currently, the plugin only allows the deployment of a single AppSync API via the appSync map:
``` yaml
custom:
  appSync:
    name: all-in-one-appsync-endpoint-yikes
    schema: AppSync/schema.graphql
    authenticationType: OPENID_CONNECT
    openIdConnectConfig:
    ...
    serviceRole: AllInOneServiceRole
    dataSources:
    ...
    mappingTemplatesLocation: ...
    mappingTemplates:
    ...
```

What Multiple APIs might look like:
```yaml
custom:
  appSync:
    - name: authenticated-appsync-endpoint
      schema: AppSync/schema.graphql # or something like AppSync/private/schema.graphql
      authenticationType: OPENID_CONNECT
      openIdConnectConfig:
      ...
      serviceRole: AuthenticatedAppSyncServiceRole
      dataSources:
      ...
      mappingTemplatesLocation: ...
      mappingTemplates:
      ...
    - name: public-appsync-endpoint
      schema: AppSync/schema.graphql # or something like AppSync/public/schema.graphql
      authenticationType: NONE # or API_KEY, you get the idea
      serviceRole: PublicAppSyncServiceRole
      dataSources:
      ...
      mappingTemplatesLocation: ...
      mappingTemplates:
      ...
```

Reasons:
1. The issue resolution [here](https://github.com/sid88in/serverless-appsync-plugin/issues/135) does not play well with the sharing of datasources and functions across multiple separate services. I would have to duplicate an entire service and all its dependencies just to create a "public" version of the API.
2. Given that the public version of the API would most likely be a subset of the private/authenticated API, it would make sense to update them together, so that we can control authentication via resolver templates, while sharing a single API schema, should we so choose.
3. Splitting authenticated/public fields to separate APIs or services defeats the purpose of GraphQL as I would now have multiple endpoints, and have to instantiate multiple GraphQL clients.

sid88in:
seems like a good idea for the plugin to be able to deploy multiple API's. Although the point of GraphQL is to have a single endpoint but in my experience I have seen organizations using multiple GraphQL endpoints for different purposed as you mentioned. 

Is that something you would like to contribute?

ZY-Ang:
@sid88in Sure. Was browsing around the codebase. Doesn't seem to be too tough of a task. How do I get started?

sid88in:
[...]

Issue #135, opened 2018-08-08
Is it possible to deploy multiple API's form one serverless.yml?


nicokruger:
As per the title, we have the requirement that we have two appsync API's - one for authenticated users, and one for non authenticated users. Much of the lambda code, resolvers, graphql types etc. is shared so it would be preferable if we could deploy from one serverless.yml.

Is there any way to deploy multiple appsync API's in one serverless stack?

nicokruger:
Ok, thanks @bboure 

Basically what I suspected, and while I'm not necessarily thrilled about it, I can understand it :)

Really this boils down to the difficulty in AppSync in doing unauthenticated users with AMAZON_COGNITO_POOLS based AppSync API's and has got nothing to do with this plugin.

The plugin works great :)

PR #155, opened 2018-09-17
Add support Multiple APIs

ZY-Ang:
This PR resolves:
- [#154 Feature Request/Discussion: Support for multiple APIs](https://github.com/sid88in/serverless-appsync-plugin/issues/154)
- [#135 Is it possible to deploy multiple API's form one serverless.yml](https://github.com/sid88in/serverless-appsync-plugin/issues/135)
- The second last paragraph of [this article](https://read.acloud.guru/deploy-a-graphql-service-on-aws-with-the-serverless-framework-7af8fc22a01d)

# What it does
Previously, deploying multiple APIs from this plugin was not possible, as the appSync plugin only allows for an `Object` configuration within the custom.appSync variable:
``` yaml
custom:
  appSync:
    name: all-in-one-appsync-endpoint-yikes
    schema: AppSync/schema.graphql
    authenticationType: OPENID_CONNECT
    openIdConnectConfig:
    ...
    serviceRole: AllInOneServiceRole
    dataSources:
    ...
    mappingTemplatesLocation: ...
    mappingTemplates:
    ...
```
This PR adds support for a variable of type `Array` to support deployment to multiple AppSync APIs.
```yaml
custom:
  appSync:
    - name: authenticated-appsync-endpoint
      schema: AppSync/schema.graphql # or something like AppSync/private/schema.graphql
      authenticationType: OPENID_CONNECT
      openIdConnectConfig:
      ...
      serviceRole: AuthenticatedAppSyncServiceRole
      dataSources:
      ...
      mappingTemplatesLocation: ...
      mappingTemplates:
      ...
    - name: public-appsync-endpoint
      schema: AppSync/schema.graphql # or something like AppSync/public/schema.graphql
      authenticationType: NONE # or API_KEY, you get the idea
      serviceRole: PublicAppSyncServiceRole
      dataSources:
      ...
      mappingTemplatesLocation: ...
      mappingTemplates:
      ...
```
# Does it break anything?
No. The old `Object` setup still works fine. If a map/ `Object` is provided in the config, the logicalId of the appsync API remains as `GraphQlApi`.
# Caveats
1. `deleteGraphQLEndpoint` functionality that helps to retain existing API key and URLs will continue to work if setups are maps but will change once names are changed, as per AWS CloudFormation docs.
2. Resources dependent on any existing roles, policies, datasources or resolvers generated by the plugin, will be updated to match Serverless' naming convention. That is,
   - `GraphQlDs<DATASOURCE>` > `<DATASOURCE>GraphQlDs`
   - `GraphQlResolver<RESOLVER>` > `<RESOLVER>GraphQlResolver`
   - `GraphQlDs<DATASOURCE>Role` > `<DATASOURCE>GraphQlDsRole`

### DIFF
--- a/index.js
+++ b/index.js
@@ -345,8 +345,8 @@ class ServerlessAppsyncPlugin {
           ]
         }
       };
-      
-      return Object.assign({}, acc, { [this.getLogicalId(ds, RESOURCE_DATASOURCE_ROLE)]: resource });
+      // NOTE: No two AppSync APIs should share datasources. For potential fine-grain access implementation.
+      return Object.assign({}, acc, { [this.getLogicalId(config, this.getLogicalId(ds, RESOURCE_DATASOURCE_ROLE))]: resource });
     }, {});
   }
   
@@ -454,7 +454,7 @@ class ServerlessAppsyncPlugin {
       if (ds.config && ds.config.serviceRoleArn) {
         resource.Properties.ServiceRoleArn = ds.config.serviceRoleArn;
       } else {
-        const dataSourceRoleLogicalId = this.getLogicalId(ds, RESOURCE_DATASOURCE_ROLE);
+        const dataSourceRoleLogicalId = this.getLogicalId(config, this.getLogicalId(ds, RESOURCE_DATASOURCE_ROLE));
         // If a Role Resource was generated for this DataSource, use it
         const role = this.serverless.service.provider.compiledCloudFormationTemplate.Resources[dataSourceRoleLogicalId];
         if (role) {
@@ -484,7 +484,8 @@ class ServerlessAppsyncPlugin {
       } else if (ds.type !== 'NONE') {
         throw new this.serverless.classes.Error(`Data Source Type not supported: '${ds.type}`);
       }
-      return Object.assign({}, acc, { [this.getLogicalId(ds, RESOURCE_DATASOURCE)]: resource });
+      // NOTE: No two AppSync APIs should share datasources. For potential fine-grain access implementation.
+      return Object.assign({}, acc, { [this.getLogicalId(config, this.getLogicalId(ds, RESOURCE_DATASOURCE))]: resource });
     }, {});
   }
 
--- a/index.js
+++ b/index.js
@@ -7,6 +7,14 @@ const runPlayground = require('./graphql-playground');
 const getConfig = require('./get-config');
 
 const MIGRATION_DOCS = 'https://github.com/sid88in/serverless-appsync-plugin/blob/master/README.md#cfn-migration';
+const RESOURCE_API = "GraphQLApi";
+const RESOURCE_API_CLOUDWATCH_LOGS_ROLE = "GraphQlApiCloudWatchLogsRole";
+const RESOURCE_API_KEY = "GraphQlApiKey";
+const RESOURCE_SCHEMA = "GraphQlSchema";
+const RESOURCE_URL = "GraphQlApiUrl";
+const RESOURCE_RESOLVER = "GraphQlResolver";
+const RESOURCE_DATASOURCE = "GraphQlDs";
+const RESOURCE_DATASOURCE_ROLE = "GraphQlDsRole";
 
 class ServerlessAppsyncPlugin {
   constructor(serverless, options) {
@@ -65,7 +73,7 @@ class ServerlessAppsyncPlugin {
         + `is no longer supported. See ${MIGRATION_DOCS} for more information`);
     };
     this.hooks = {
-      'before:deploy:initialize': () => this.validateSchema(),
+      'before:deploy:initialize': () => this.validateSchemas(),
       'delete-appsync:delete': () => this.deleteGraphQLEndpoint(),
       'graphql-playground:run': () => this.runGraphqlPlayground(),
       'deploy-appsync:deploy': generateMigrationErrorMessage('deploy-appsync'),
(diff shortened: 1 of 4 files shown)


index.js
```diff
@@ -97,78 +105,95 @@ class ServerlessAppsyncPlugin {
...
+    const errors = asts.reduce((accumulatedErrors, currentAst) => {
+      const currentErrors = validateSchema(currentAst);
+      if (!currentErrors.length) {
+        return accumulatedErrors;
+      } else if (!accumulatedErrors.length) {
+        return currentErrors;
+      } else {
+        return accumulatedErrors.concat(currentErrors || []);
+      }
+    });
```
cbaron: Can you provide an initialValue to the `reduce` method?

According to the [documentation](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/Reduce) :

```
If no initial value is supplied, the first element in the array will be used.
```

The following case 

```
else if (!accumulatedErrors.length)
```

will never occur, and `accumulatedErrors` will include `asts[0]`

index.js
```diff
@@ -497,10 +525,23 @@ class ServerlessAppsyncPlugin {
...
 
-  getGraphQlApiOutputs() {
+  getLogicalId(config, resourceType) {
+    // Similar to serverless' implementation of functions
+    // (e.g. getUser becomes GetUserLambdaFunction for CloudFormation logical ID,
+    //  myService becomes MyServiceGraphQLApi or `MyService${resourceType}`)
+    if (config.isSingleConfig) {
+      // This will ensure people with CloudFormation stack dependencies on the previous
+      // version of the plugin doesn't break their {@code deleteGraphQLEndpoint} functionality
+        return this.getCfnName(resourceType);
```
cbaron: 👍

index.js
```diff
@@ -309,6 +333,7 @@ class ServerlessAppsyncPlugin {
           },
           Policies: [
             {
+              // TODO: Why are inline policies getting fine-grained policy names? There is no benefit to this.
               PolicyName: this.getDataSourceCfnName(ds.name) + "Policy",
```
ZY-Ang: @sid88in @cbaron Is there any reason to the above I'm missing? Inline policies that aren't a Policy resource don't really have a requirement to have datsource specific names, being attached to only one parent role.

index.js
ZY-Ang: Good point. I set the initial value to an empty array and removed the accumulatedErrors base case and updated get-config to reflect the changes. If an empty configuration (i.e. plugin is enabled but nothing in config) exists, get-config will return an empty array which can be used to call `.map` on.

index.js
cbaron: @ZY-Ang, I don't have much context here.  Its a good question.  My guess is that @sid88in wanted to ensure descriptive policy names.

bboure: @ZY-Ang @cbaron I wrote that line. 
I am not sure I understood your question though...
`PolicyName` is a required field in cloudformation
https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-iam-policy.html#cfn-iam-policy-policyname
I just gave it a name related to the Datasource it refers to as a reference.

Or did I miss something?

ZY-Ang: @bboure I was just confused. It's like arguing `pineappleRole.penPolicy` vs `pineappleRole.penPineapplePolicy`. Although the latter would make sense if it was a policy resource and not an inline policy

bboure: @ZY-Ang I am ok with changing it. I used this for no particular reason other than it was required.  What do you suggest?  Something like `defaultPolicy` or `root` (Like shown [here](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-iam-role.html)) or something?

pmuens:
@bboure @ZY-Ang @sid88in Just glimpsed over the PR.

Yes, usually renaming will delete stack resources if you don't specify a `DeletionPolicy` of `retain` (if possible).

Haven't digged deeper, but it would be nice to understand (from a high-level) what this PR actually does / should do. Buy understanding is that it should spin-up a whole new, separate API which lives side-by-side with the old one 🤔

bboure:
@pmuens Yeah, well that was my point here. I would indeed expect resources to be recreated after upgrading and redeploying or fail if you have a `DeletionPolicy` policy.
Not sure what impact this could have and that will probably depend on each individual use case. 
If you have references to the old url somewhere (e.g: Cloudfront), things could stop working...

sid88in:
thanks @bboure @pmuens for your inputs. @ZY-Ang I am reverting this change for now till we measure E2E impact and test backward compatibility. I don't want anyone's resources to be dropped :)

https://github.com/sid88in/serverless-appsync-plugin/pull/156

looking into npm unpublish

sid88in:
@ZY-Ang unfortunately I can't re-open merged PR. So, please create a new one when you get time.

thanks
Sid

ZY-Ang:
@bboure @pmuens CloudFormation logicalIds for the primary GraphQlApi do not change only if the existing setup remains the same (I.e. map). All other sub resources' logicalIds like data sources or resolvers change. It is a good idea to have more apis live alongside but there is no way to differentiate them easily when converted into a Array, unless you have some additional rule system that says hey, this API logical ID should be GraphQlApi

cbaron:
I think the simplest solution would be the introduction of an additional config,

```logicalIdVersion```

which, when omitted would handle the current naming conventions (backwards compatible), and when included, a value of `1` would reference the new conventions created by @ZY-Ang.

PR #161, opened 2018-10-09
Remove unused `accountId` parameter

alexjurkiewicz:
Everywhere that needed an account ID was using the AWS::AccountId
pseudo-parameter [1] instead.

[1] https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/pseudo-parameter-reference.html

### DIFF
--- a/README.md
+++ b/README.md
@@ -66,7 +66,6 @@ Add the following example config to the custom section of ```serverless.yml```
 
 ```yaml
 custom:
-  accountId: abc # found here https://console.aws.amazon.com/billing/home?#/account
   appSync:
     name:  # defaults to api
     # apiKey # only required for update-appsync/delete-appsync
--- a/index.js
+++ b/index.js
@@ -247,10 +247,9 @@ class ServerlessAppsyncPlugin {
                     "Resource": [
                       {
                         "Fn::Sub" : [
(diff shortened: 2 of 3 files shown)


bboure:
In fact `accountId` was still in the doc but unused. 
In policy references, `AWS::AccountId` was used through the `Fn:Sub` helper.
I am not sure (but I might be wrong) that it works the way you changed it.

By the way, we should also remove `config.region` and replace it with `AWS::Region` I guess.

I'll test all this when I have a moment.
Thanks

alexjurkiewicz:
`Fn::Sub` and `Ref` support the same references, Sub just allows interpolation into a string.

I will update the PR to drop region too. I thought it might be there for cross-region support, but looking closer now I see it's not.

bboure:
@alexjurkiewicz Awesome thank you 🎢

alexjurkiewicz:
I've added a commit to remove region as well. To be honest, I'm not sure if the code for that part is good. I guessed at how serverless works... I know CF not serverless API :D

alexjurkiewicz:
Sorry for the delay. Turns out I don't have the time to remove region as well, so I've removed that commit. We're back to where we started, just removing the `accountId` param for now 👍

Issue #144, opened 2018-08-19
How to overcome Number of resources is greater than maximum allowed, 200


harleyguru:
Hi @sid88in 
I've been using this lovely plugin for a long time.

Recently, I am experiencing one annoying troubles to deploy as my backend is getting bigger and bigger day by day.

Since my AppSync service includes a lot of data sources(backed DynamoDB tables) and templates, I am getting this resource number limit issue.

I am wondering how we can split the AppSync service into multiple stacks.
Of course, I've been trying to use serverless-plugin-additional-stacks or serverless-plugin-split-stacks etc. But they won't work with this plugin.

Can you guide me for this issue?
Thanks

bboure:
@harleyguru 
Yeah, the problem with appsync is that we cannot easily split it into several services like you would with Api gateway because, of course, there is only one endpoint.

I see 2 ways of doing this:

1) We change the plugin to be able to create data sources on an existing Appsync Resource.
Basically by allowing overriding [this](https://github.com/sid88in/serverless-appsync-plugin/blob/master/index.js#L164) and skipping [this](https://github.com/sid88in/serverless-appsync-plugin/blob/master/index.js#L116) on sub-stacks.
That way, you could have one "master" `serverless.yml` that defines the appsync resource (Schema, Authentication, etc.) + a bunch of other `serverless.yml` files that just creates data-sources and attaches them to the API. 

2) Use [Nested Stacks](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/using-cfn-nested-stacks.html) somehow...
In one single `serverless.yml`, we could maybe either detect the limit is reached (not sure how) and split automatically, or allow marking each Datasource as being part of a group (sub-stack)...

On the other hand, I don't know how you built your stack but you could start by moving some non-related resources in another stack: e.g: DynamoDb tables, Cognito User Pool, etc. You can still reference to them in your appsync stack using [`Outputs`](https://serverless.com/framework/docs/providers/aws/guide/variables#reference-cloudformation-outputs)

harleyguru:
Hello @bboure ,
First off, Thanks for your quick reply!

Yeah, I guess nested stacks will be a good solution for this problem.
The main point of growing number of resources is in a lot of mapping-templates.
In CFN console, I see a lot of template resources actually.

The question is if serverless-plugin-split-stacks(this plugin splits the stack into multiple nested stacks) will work for AppSync templates resources.

Thanks

[...]

PR #171, opened 2018-11-07
Use a different hook for adding resources to allow sls package and the use of other plugins (e.g. split stacks)

kinyat:
We got the same issue as https://github.com/sid88in/serverless-appsync-plugin/issues/144

After doing some investigations, we found that by changing the hook event in appsync plugin, it will allow the [split stacks plugin](https://github.com/dougmoscrop/serverless-plugin-split-stacks) to work

`stacks-map.js` used by the split stacks plugin

```js
module.exports = {
  'AWS::AppSync::ApiKey': { destination: 'AppSync', allowSuffix: true },
  'AWS::AppSync::DataSource': { destination: 'AppSync', allowSuffix: true },
  'AWS::AppSync::GraphQLApi': { destination: 'AppSync', allowSuffix: true },
  'AWS::AppSync::GraphQLSchema': { destination: 'AppSync', allowSuffix: true },
  'AWS::AppSync::Resolver': { destination: 'AppSync', allowSuffix: true }
}
```

image 4

After deploy

image

We also tested the deployed appsync and it all works perfectly.

One more benefit of this change is that doing `serverless package` will now include the resource added by the AppSync plugin so it is easier to debug.

Some generated AppSync resources in `cloudformation-template-update-stack.json` after running `serverless package`.

image

### DIFF
--- a/index.js
+++ b/index.js
@@ -72,7 +72,7 @@ class ServerlessAppsyncPlugin {
       'graphql-playground:run': () => this.runGraphqlPlayground(),
       'deploy-appsync:deploy': generateMigrationErrorMessage('deploy-appsync'),
       'update-appsync:update': generateMigrationErrorMessage('update-appsync'),
-      'before:deploy:deploy': () => this.addResources(),
+      'after:aws:package:finalize:mergeCustomProviderResources': () => this.addResources(),
     };
   }
 
--- a/README.md
+++ b/README.md
@@ -28,6 +28,9 @@ Tired of 🚀 **deploying**, ✏️ **updating**, and ❌ **deleting** your AppS
 * [Getting Started](#-getting-started)
 * [Installation](#-installation)
 * [Usage](#️-usage)
+* [Notes](#-notes)
+    * [Offline Support](#offline-support)
+    * [Split Stacks Plugin](#split-stacks-plugin)
 * [Contributing](#-contributing)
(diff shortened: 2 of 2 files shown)


index.js
```diff
@@ -72,7 +72,7 @@ class ServerlessAppsyncPlugin {
       'graphql-playground:run': () => this.runGraphqlPlayground(),
       'deploy-appsync:deploy': generateMigrationErrorMessage('deploy-appsync'),
       'update-appsync:update': generateMigrationErrorMessage('update-appsync'),
-      'before:deploy:deploy': () => this.addResources(),
+      'after:aws:package:finalize:mergeCustomProviderResources': () => this.addResources(),
```
bboure: @kinyat I am not familiar with this hook but I see it is also used by `serverless-plugin-split-stacks`
Does this means that `serverless-plugin-split-stacks` has to be placed **after** `serverless-appsync-plugin` in serverless.yml file plugins list? (so that first this plugin adds teh resource, and then the other plugin splits them).
If so, this could be specified somewhere in the readme...

kinyat: Cool. I will update the readme for the usage with the split stacks plugin

kinyat: I have updated the readme.md

sid88in:
thanks @kinyat and @bboure merging 🎉

PR #176, opened 2018-11-13
Fix RESOURCE_API typo

ZY-Ang:
Should be GraphQlApi, not GraphQLApi


sid88in:
@ZY-Ang is this a critical fix where we need to create a new version or can this wait without any blocker?

bboure:
@sid88in I think this breaks backward compatibility. So it should be released asap
@ZY-Ang  ?

ZY-Ang:
Yeah. My bad.

sid88in:
ok doing another version. thanks guys

PR #180, opened 2018-11-25
Fix playground

hardchor:
Hey! I'm not sure how this ever worked, but async serverless hooks expect a Promise back. I don't really like the "empty Promise hack" to stop if from terminating immediately, but I also couldn't come up with a better solution.

I also updated the call signature of `provider.request()` to stop it from throwing `Serverless: WARNING: Inappropriate call of provider.request()` in debug mode, and noticed that the `yarn.lock` was out of sync.

### Before:
```bash
Serverless: Invoke graphql-playground
Serverless: WARNING: Inappropriate call of provider.request()
Serverless: WARNING: Inappropriate call of provider.request()
Serverless: WARNING: Inappropriate call of provider.request()
✨  Done in 2.36s.
```

### After:
```bash
Serverless: Invoke graphql-playground
  koa:application use voyager +0ms
  koa:application listen +1ms
Serverless: Graphql Playground Server Running at: http://localhost:3000
```

### DIFF
--- a/index.js
+++ b/index.js
@@ -154,9 +154,11 @@ class ServerlessAppsyncPlugin {
   runGraphqlPlayground() {
     // Use the first config or config map
     const firstConfig = this.loadConfig()[0];
-    runPlayground(this.serverless.service, this.provider, firstConfig, this.options).then((url) => {
-      this.serverless.cli.log(`Graphql Playground Server Running at: ${url}`);
-    });
+    return runPlayground(this.serverless.service, this.provider, firstConfig, this.options)
+      .then((url) => {
+        this.serverless.cli.log(`Graphql Playground Server Running at: ${url}`);
+      })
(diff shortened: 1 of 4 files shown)


index.js
```diff
@@ -154,9 +154,11 @@ class ServerlessAppsyncPlugin {
   runGraphqlPlayground() {
     // Use the first config or config map
     const firstConfig = this.loadConfig()[0];
-    runPlayground(this.serverless.service, this.provider, firstConfig, this.options).then((url) => {
-      this.serverless.cli.log(`Graphql Playground Server Running at: ${url}`);
-    });
+    return runPlayground(this.serverless.service, this.provider, firstConfig, this.options)
```
hardchor: without the return (and the empty promise™️), the serverless hook will immediately terminate

sid88in:
thanks @hardchor

PR #182, opened 2018-11-27
Allow for nested Mapping Templates

lkhari:
In order to reduce the large file containing all mapping templates I have made this change to allow multiple file inputs. 

I am flattening the input mapping templates into a single array. 
This is tested with 

- array of multiple files
- array of files and mappingtemplates
- nothing
- array of mappingtemplates

I don't believe it will work with files in files though.

Usage example
``` 
  appSync:
    mappingTemplates:
      - ${file(mapping-templates/queries.yml)}
      - ${file(mapping-templates/mutation.yml)}
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -95,6 +95,7 @@ custom:
         field: getUserInfo
         request: # request mapping template name
         response: # response mapping template name
+      - ${file({fileLocation}.yml)} # link to a file with arrays of mapping templates
     schema: # defaults schema.graphql
     dataSources:
       - type: AMAZON_DYNAMODB
--- a/index.js
+++ b/index.js
@@ -509,7 +509,8 @@ class ServerlessAppsyncPlugin {
     };
   }
   getFunctionConfigurationResources(config) {
-    return config.functionConfigurations.reduce((acc, tpl) => {
+    const flattenedFunctionConfigurationResources = config.functionConfigurations.reduce((accumulator, currentValue) => accumulator.concat(currentValue), []);
+    return flattenedFunctionConfigurationResources.reduce((acc, tpl) => {
       const reqTemplPath = path.join(config.mappingTemplatesLocation, tpl.request);
       const respTemplPath = path.join(config.mappingTemplatesLocation, tpl.response);
       const requestTemplate = fs.readFileSync(reqTemplPath, 'utf8');
--- a/index.js
+++ b/index.js
@@ -509,7 +509,8 @@ class ServerlessAppsyncPlugin {
   }
 
   getResolverResources(config) {
-    return config.mappingTemplates.reduce((acc, tpl) => {
+    const flattenedMappingTemplates = config.mappingTemplates.reduce((accumulator, currentValue) => accumulator.concat(currentValue), []);
+    return flattenedMappingTemplates.reduce((acc, tpl) => {
       const reqTemplPath = path.join(config.mappingTemplatesLocation, tpl.request);
       const respTemplPath = path.join(config.mappingTemplatesLocation, tpl.response);
       const requestTemplate = fs.readFileSync(reqTemplPath, 'utf8');


sid88in:
@lkhari thanks! 

1. Are these changes backward compatible?
2. Can you please update readme somewhere to mention this feature.

bboure:
Thanks @lkhari, change looks good, however, I wonder if Serverless already supports something like that natively.

The following might work without changing anything:

````yml
appSync:
  - ${file(mapping-templates/queries.yml)}
  - ${file(mapping-templates/mutation.yml)}
````

Then, inside queries.yml and mutation.yml you should have

````yml
  mappingTemplates:
    - dataSource: query1
````

````yml
  mappingTemplates:
    - dataSource: mutation1
````

It is not anywhere in serverless doc though 😄 

Check this: https://stackoverflow.com/questions/47833344/declaring-resources-in-multiple-files-in-serverless-framework post.

**Comment by lkhari** [comment]
@sid88in Yup it's backward compatible.  Added it to the readme.

hey @bboure I did try it out what you suggested and it didn't work. I believe the reason for it is that you are applying an array of objects to Appsync, whereas we would want to keep Appsync still as an object. 

If we were to remove the hypens and have multiple mapping templates you would cause a duplicate mapping key error.

**Comment by bboure** [comment]
@lkhari OK thanks for trying. It must work differently internally for Resource then in serverless.

**Comment by bboure** [comment]
@lkhari Would it be easily possible to also allow something like

````yml
appsync:
  - ${file(mapping-templates/resources1.yml)}
  - ${file(mapping-templates/resources2.yml)}
````

where both file would have `mappingTemplates` and `dataSources`.
That way we could group what goes together in the same file.

**Comment by Foosballfan** [comment]
@bboure The above syntax would not work as it resolves to appsync being an array:
```
appsync: 
  [
    {
      mappingTemplates: 
        [
          {
            ...
          }
        ]  
    },
    {
      mappingTemplates: 
        [
          {
            ...
          }
        ]  
    }
  ] 
```
When what it should be is an object:
```
appsync: 
    {
      name: 'name',
      mappingTemplates: 
        [
          {
            ...
          }
        ]  
    }
```

@lkhari is proposing a change which allows us to pass an array of arrays for mapping templates and then flattening them into a single array which is what the library requires but is not possible with yaml.

```
appsync: 
    {
      name: 'name',
      mappingTemplates: 
        [
          [
            { ... },
            { ... },
            { ... }
          ],
          [
            { ... },
            { ... },
            { ... }
          ]
        ]  
    }
```
becomes
```
appsync: 
    {
      name: 'name',
      mappingTemplates: 
        [
            { ... },
            { ... },
            { ... },
            { ... },
            { ... },
            { ... }
        ]  
    }
```

**Comment by bboure** [comment]
@Foosballfan what I mean is that the plugin could check if `appsync` is an object or an array of objects beforehand.
If `appsync` is an object, keep same logic as now.
If `appsync` is an array of objects, merge them first, and keep going the normal workflow.

**Comment by lkhari** [comment]
@bboure I assume you are talking about mappingtemplates property inside appsync instead of appsync. Since appsync is always an object.

That could work, but if you have a mixture of arrays and objects, it would have to go through the whole array to see if the array contains only objects or arrays. Which wouldn't add any improvements from what I can think of.

**Comment by vinicius-ronconi** [comment]
Does it work only for mappingTemplates?
I'm trying to do the same with the dataSources, but it is not working.

My serverless.yml looks like:
```
custom:
  appSync:
    # ...
    dataSources:
      - ${file(./src/dataSources/ds1.yaml)}
      - ${file(./src/dataSources/ds2.yaml)}
```

And my /src/dataSources/*.yaml files look like:
```
dataSources:
  - type: AMAZON_DYNAMODB
    name: "myAwesomeDataSource"
    config:
      tableName: { Ref: myAwesomeTable }
```

When I deploy, I get the following error message:
```Serverless: Packaging service...

  Serverless Error ---------------------------------------

  Data Source Type not supported: 'undefined

  Get Support --------------------------------------------
     Docs:          docs.serverless.com
     Bugs:          github.com/serverless/serverless/issues
     Issues:        forum.serverless.com

  Your Environment Information -----------------------------
     OS:                     darwin
     Node Version:           11.0.0
     Serverless Version:     1.36.1
```


But If I do something similar with the mappingTemplate or move the dataSource declaration back to the serverless.yml file, the stack is created correctly.

lkhari:
> Does it work only for mappingTemplates?
> I'm trying to do the same with the dataSources, but it is not working.
> 
> My serverless.yml looks like:
> 
> ```
> custom:
>   appSync:
>     # ...
>     dataSources:
>       - ${file(./src/dataSources/ds1.yaml)}
>       - ${file(./src/dataSources/ds2.yaml)}
> ```
> And my /src/dataSources/*.yaml files look like:
> 
> ```
> dataSources:
>   - type: AMAZON_DYNAMODB
>     name: "myAwesomeDataSource"
>     config:
>       tableName: { Ref: myAwesomeTable }
> ```
> When I deploy, I get the following error message:
> 
> ```
> 
>   Serverless Error ---------------------------------------
> 
>   Data Source Type not supported: 'undefined
> 
>   Get Support --------------------------------------------
>      Docs:          docs.serverless.com
>      Bugs:          github.com/serverless/serverless/issues
>      Issues:        forum.serverless.com
> 
>   Your Environment Information -----------------------------
>      OS:                     darwin
>      Node Version:           11.0.0
>      Serverless Version:     1.36.1
> ```
> But If I do something similar with the mappingTemplate or move the dataSource declaration back to the serverless.yml file, the stack is created correctly.

Yea it only works for mapping templates @vinicius-ronconi. Sorry for the late reply, i'm travelling right now. I might be able to look into it but to be honest how many datasources do you have? Surely just one file is good enough?

vinicius-ronconi:
Hi @lkhari !

No problem at all. I'm new to AppSync and I was just wondering if I was doing something wrong! :)

You're right, I have just a few datasources, so keep them on the main serverless.yml file won't be a problem.

lkhari:
The `-` denotes it's part of an array. So with what you wrote above it, once flattened and as a json, would look like:

```
appSync: {
  dataSources: [
    {
      dataSources: [
        {
          type: AMAZON_DYNAMODB,
          name: "myAwesomeDataSource",
          config: {
            tableName: { Ref: myAwesomeTable }
          }
        }
      ]
    },    {
      // what ever is in file two
    }
  ]
}
```

If you want one file outside but multiples sources you can write it like this:
Serverless yml:
```
dataSources:
      - ${file(./src/dataSources/ds1.yaml)}
```

./src/dataSources/ds1.yaml:
``` 
- type: AMAZON_DYNAMODB
  name: "myAwesomeDataSource"
  config:
  tableName: { Ref: myAwesomeTable }
- type: AMAZON_DYNAMODB
  name: "myOtherAwesomeDataSource"
  config:
  tableName: { Ref: myOtherAwesomeTable }
```
Which turns into 
```
appSync: {
  dataSources: [
    {
      type: AMAZON_DYNAMODB,
      name: "myAwesomeDataSource",
      config: {
        tableName: { Ref: myAwesomeTable }
      }
    },    {
      type: AMAZON_DYNAMODB,
      name: "myOtherAwesomeDataSource",
      config: {
        tableName: { Ref: myOtherAwesomeTable  }
      }
    }
  ];
}
```

Notice there dataSources array in the second example doesn't hold an object which contains another datasource array. This is the main reason why what you wrote couldn't of worked.
 
<details>
 <summary>Or you can write multiple files but only contain one datasource per file like: 
</summary>

serverless.yml
```
dataSources:
      - ${file(./src/dataSources/ds1.yaml)}
      - ${file(./src/dataSources/ds2.yaml)}
```

./src/dataSources/ds1.yaml:
```
type: AMAZON_DYNAMODB
name: "myAwesomeDataSource"
config:
tableName: { Ref: myAwesomeTable }
```
./src/dataSources/ds2.yaml:
```
type: AMAZON_DYNAMODB
name: "myOtherAwesomeDataSource"
config:
tableName: { Ref: myOtherAwesomeTable  }
```
</details>


 
Hope this helps @vinicius-ronconi

vinicius-ronconi:
Awesome @lkhari ! Thank you!!

Issue #197, opened 2019-01-08
Cognito's userPoolConfig.defaultAction is a required property


alexleonescalera:
When you don't include the `userPoolConfig.defaultAction` property (ALLOW or DENY), it is not generated for the CloudFormation JSON. This leads to the stack creation failing and an error such as this:

```bash
CloudFormation - CREATE_FAILED - AWS::AppSync::GraphQLApi - MyGraphQlApi
```

```bash
Serverless Error ---------------------------------------

An error occurred: MyGraphQlApi - Invalid default effect type. (Service: AWSAppSync; Status Code: 400; Error Code: BadRequestException; Request ID: ...).
```

Unfortunately, this most likely is due to an error in the [CloudFormation documentation](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-appsync-graphqlapi-userpoolconfig.html#cfn-appsync-graphqlapi-userpoolconfig-defaultaction) for `userPoolConfig`, which states (as of the time of this ticket's creation) that `DefaultAction` is not required. This property is required, as seen [AppSync API reference doc](https://docs.aws.amazon.com/appsync/latest/APIReference/API_UserPoolConfig.html). The documentation issue has been reported.

I'll provide a pull request in an attempt to fix this.

PR #198, opened 2019-01-08
Cognito's userPoolConfig.defaultAction is a required property

alexleonescalera:
Solves https://github.com/sid88in/serverless-appsync-plugin/issues/197


sid88in:
thanks @alexleonescalera

PR #206, opened 2019-01-26
Add default filename conventions for mappingTemplates

caki0915:
# Problem
When specifying mapping-templates in `serverless.yml`, the file gets very big very fast.
In most cases, the user wants to use the same structure and naming conventions for all requests and response template. 

# Solution 
If request and response is omitted a default naming convention will be used instead:
`{Type}.{Field}.{Request}.vtl`  and `{Type}.{Field}.{Resonse}.vtl`, This will make the serverless file smaller.

# Example
These fields in `serverless.yml`:

```
- field: getSomeData
   dataSource: myDatasource
   type: Query

- field: putSomeData
   dataSource: myDatasource
   type: Mutation

- field: getOtherData
   dataSource: myDatasource
   type: Query

```

Will map by default to:

```
Query.getSomeData.request.vtl
Query.getSomeData.response.vtl

Mutation.putSomeData.request.vtl
Mutation.outSomeData.response.vtl

Query.getOtherData.request.vtl
Query.getOtherData.response.vtl
```

### DIFF
--- a/index.js
+++ b/index.js
@@ -522,8 +522,8 @@ class ServerlessAppsyncPlugin {
   getFunctionConfigurationResources(config) {
     const flattenedFunctionConfigurationResources = config.functionConfigurations.reduce((accumulator, currentValue) => accumulator.concat(currentValue), []);
     return flattenedFunctionConfigurationResources.reduce((acc, tpl) => {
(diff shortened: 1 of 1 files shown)


caki0915:
This pull request fixes #207

sid88in:
thanks @caki0915 I am assuming you tested this out.

PR #216, opened 2019-02-25
Feature/support datasource arrays

bboure:
Just as mapping templates in multiple files are supported, dataSources are now supported as well.

In each file, the following will work:

As an Object whose keys are the datasource names:

````yml
# file1.yml
dataSourceName:
  type: AWS_LAMBDA
  config:
    functionName: myFunction
````

As an Array, specifying `name`
````yml
# file2.ylm
- type: AWS_LAMBDA
  name: dataSourceName2
  config:
    functionName: myFunction2
````

````yml
datasources:
 - ${file(file1.yml)}
 - ${file(file2.yml)}
````

Passing an array or object directly will keep working as well:

````yml
datasources:
  - type: AWS_LAMBDA
    name: dataSourceName
    config:
      functionName: myFunction
````

````yml
datasources:
  dataSourceName2:
    type: AWS_LAMBDA
    name: dataSourceName2
    config:
      functionName: myFunction2
````

### DIFF
--- a/README.md
+++ b/README.md
@@ -143,6 +143,7 @@ custom:
         description: 'Http endpoint'
         config:
           endpoint: # required # "https://{DOMAIN}/{PATH}"
+      - ${file({dataSources}.yml)} # link to a file with an array or object of datasources
     substitutions: # allows to pass variables from here to velocity templates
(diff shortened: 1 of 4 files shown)


bboure:
@lkhari since you did something similar, a review would be welcome.

PR #220, opened 2019-02-28
Added automatic CloudWatch LogGroup creation & deletion with expiry

sebflipper:
Currently if you enable logging an AWS role is created with permissions to create a new log group, but because this log group has not been created by serverless as part of the CloudFormation it is not cleaned up/deleted when removing the project.

This will automatically add a log group for AppSync if a `logConfig.level` has been set and also set a log expiry based on [`logRetentionInDays`](https://serverless.com/framework/docs/providers/aws/guide/serverless.yml/) (if set).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,4 @@
 language: node_js
 node_js:
-  - "6.10"
-  - "8.10"
+  - "8"
+  - "10"
--- a/README.md
+++ b/README.md
@@ -26,6 +26,7 @@ Tired of 🚀 **deploying**, ✏️ **updating**, and ❌ **deleting** your AppS
  <summary><strong>Table of Contents</strong> (click to expand)</summary>
 
 * [Getting Started](#-getting-started)
+* [Minimum requirements](#-minimum-requirements)
 * [Installation](#-installation)
 * [Usage](#️-usage)
 * [Notes](#-notes)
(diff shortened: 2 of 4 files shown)


bboure:
Nice!. Love it.
Tests fail for node 6 because of the `...` operator; but since node 6 is almost EOL, I guess we could ditch that (and Add tests for node 10)

sebflipper:
@bboure I've removed Node v6 as it will be EOL next month and I've added the latest LTS Node v10 in 7d0190e.

bboure:
Thanks. 
Can we add a note in the readme concerning the minimum requirements? 
- node 8
- serverless XXX (not sure which version is minimum, would need to check) 
- others?

sebflipper:
@bboure I've added Node and Serverless minimum requirements in 53384bb. The minimum Serverless version I've taken from the [`package.json`](https://github.com/sid88in/serverless-appsync-plugin/blob/master/package.json#L18).

bboure:
Makes sense.

sid88in:
looks good. thanks

alexjurkiewicz:
We are running into errors deploying updates to stacks that I believe are caused by this PR.

When updating a stack that was deployed with an earlier version of the serverless-appsync-plugin without this merged, deployment fails. Example CloudFormation error:
```
GraphQlApiLogGroup AWS::Logs::LogGroup /aws/appsync/apis/j7p2mxyz already exists
```

This is because the log group was previously created implicitly in the previous configuration.

There are a few workarounds:

1. Delete the log group and deploy the stack again. This works, but isn't viable for a stack deployed in production, since you will break logging and will have to race the automatic re-creation of the log group.
2. Change this plugin to only add this resource to the catalog if you aren't migrating from an older stack version (eg, add an option to disable creating the log group)

Any comments?

sebflipper:
A 3rd workaround if you're not able to fix it via option 1, would be to:

3. Setup [Blue-Green deployment](https://aws.amazon.com/quickstart/architecture/blue-green-deployment/) 
    1. Bring up a new environment (which will create the log group).
    2. Once it's up switch traffic over to it.
    3. Then when all traffic has switched over spin down/destroy the old environment.

jplock:
I just ran into the same issue @alexjurkiewicz reported.  I'm going to try to do #1, but some additional testing around this change would have been appreciated and preferred.

bboure:
@jplock  let us know if it works for you.
@sebflipper could you help maybe with this issue? It looks like it is affecting more people.
I like the blue-green deployment strategy, and personally, this is certainly what I would do in production, no matter what; but here we might have introduced a bug worth fixing. 

Maybe option 2. suggested by @alexjurkiewicz is worth it.

jplock:
A blue-green deploy wouldn’t be ideal for me since the AppSync URL would change and I’d need to update CloudFront (15+ minutes)

bboure:
@sebflipper you can use CloudFront or Route53 to use you own domain and make it point to your AppSync endpoint. 
Then, it's just a config to change.

PR #222, opened 2019-03-03
Fix use of functionConfigurationsLocation

vlesierse:
Within my project I would like to seperate my mapping templates from my function templates by specifying different values in the `mappingTemplatesLocation` and `functionConfigurationsLocation`. As it turns out `functionConfigurationsLocation` isn't used at, which results is Serverless looking for my files in the `mappingTemplatesLocation`.

This PR will fix this behaviour. However, for backwards compatibility I have defaulted `functionConfigurationsLocation` to the same location as `mappingTemplatesLocation`.


sid88in:
looks good. thanks

PR #224, opened 2019-03-07
fixing error printout

EsrefDurna:
before this change error was
ServerlessError: Data Source Type not supported: 'HTTP

after this change it will be
ServerlessError: Data Source Type not supported: HTTP


bboure:
Thanks @EsrefDurna

Issue #232, opened 2019-03-20
Error: Relational database type is not supported


harleyguru:
I've followed the exact steps to link RDS DB cluster as a data source.
But I am getting that it's not supported type as you can see here:
![Screen Shot 2019-03-20 at 6 48 44 PM](https://user-images.githubusercontent.com/35247645/54704176-ed241380-4b42-11e9-93ee-b70f8e1e56f5.png)

This is my template:
![Screen Shot 2019-03-20 at 6 48 54 PM](https://user-images.githubusercontent.com/35247645/54704200-f7dea880-4b42-11e9-9258-41c080615de2.png)

harleyguru:
oh sorry, I didn't upgrade plugin.

harleyguru:
Hi @sid88in I've deployed successfully for RDS data source, but in the console I could notice AppSync set nothing for RDS cluster Arn.

Instead of dbClusterIdentifier, should I set DB Cluster Arn?

roznalex:
@harleyguru does it work with the DB Cluster Arn instead?

harleyguru:
hi @roznalex didn't try it yet.

roznalex:
@harleyguru fixed the issue in the [PR](https://github.com/sid88in/serverless-appsync-plugin/pull/234). After the PR will be merged, it should work fine with the "dbClusterIdentifier" config.

harleyguru:
Thanks @roznalex Could you let me know when it's merged and ready to use please?

tracstarr:
So I've been trying recently to get this to work. Using  _DbClusterIdentifier_ with the ref as defined in the docs  does not correctly set the RDS Cluster ARN. I am forced to manually modify the cluster ARN via console dropdown. 

I am using  "serverless-appsync-plugin": "^1.1.1". The _DbClusterIdentifier_ also doesn't set the ARN correctly, it only uses the ClusterName. I had to use

_dbClusterIdentifier: "arn:aws:rds:#{AWS::Region}:#{AWS::AccountId}:cluster:${self:service.name}-${self:custom.stage}"_

My ClusterDefinition

 AuroraRDSCluster:
      Type: "AWS::RDS::DBCluster"
      Properties:
        MasterUsername: ${self:custom.AURORA.USERNAME}
        MasterUserPassword: ${self:custom.AURORA.PASSWORD}
        DBSubnetGroupName:
          Ref: AuroraSubnetGroup
        Engine: aurora
[...]

PR #234, opened 2019-03-27
Bugfix/change config for relational database ds

roznalex:
Hi, @bboure. Even AWS documentation may be wrong :)
There is a fix for the [issue](https://github.com/sid88in/serverless-appsync-plugin/issues/232). I have added an automatic "dbClusterArn" generation. Now it should work correctly.


bboure:
So, doc is wrong? https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-appsync-datasource-relationaldatabaseconfig-rdshttpendpointconfig.html#cfn-appsync-datasource-relationaldatabaseconfig-rdshttpendpointconfig-dbclusteridentifier

`DbClusterIdentifier`  expects an arn?

roznalex:
> So, doc is wrong? https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-appsync-datasource-relationaldatabaseconfig-rdshttpendpointconfig.html#cfn-appsync-datasource-relationaldatabaseconfig-rdshttpendpointconfig-dbclusteridentifier
> 
> `DbClusterIdentifier` expects an arn?

@bboure yes. I tested this behaviour within my environment. It only works if you provide full cluster ARN for a cluster.

index.js
```diff
@@ -104,6 +104,31 @@ class ServerlessAppsyncPlugin {
     return { "Fn::GetAtt": [lambdaLogicalId, "Arn"] };
   }
 
+  getDbClusterArn(config) {
+    if (config && config.dbClusterIdentifier) {
+      return this.generateDbClusterArn(config.dbClusterIdentifier, config.region);
```
bboure: @roznalex Could we maybe implement it so that:
- if an arn is given, use it
- if an "identifier" is given, generate the Arn from it

I would also keep an eye on Cloudformation's doc. I wouldn't be surprised if they'd "fix it" or add a new key `dbClusterArn`

roznalex: @bboure sure. So should we add support for two options: dbClusterIdentifier and dbClusterArn?

bboure: No, I would wait and see what AWS does.
On second thought, let's leave it as it is and keep it "in sync" with cloudformations

bboure: @roznalex ok, so in fact that means that `dbClusterIdentifier` should receive an arn, and we could forget about generating it
What do you think?

roznalex: @bboure I'm not sure that it is the right decision, because a user needs to build this ARN by himself. According to [documentation](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-rds-dbcluster.html) Ref function returns a resource name.

bboure: There is the `Fn::GetAtt` function, but now I see that it does not return the Arn (not sure).
Let's keep it as it is. It would be good to maybe write a post on AWS forum and ask for more info on this.

bboure:
@roznalex A little follow up on this. 
Sorry for not merging this earlier. I really want to merge this asap, but I am a bit afraid that this could break again in the future. 
I opened a question on the [AWS forum](https://forums.aws.amazon.com/thread.jspa?threadID=301588). I'd like to have feedback from them before doing anything.

sid88in:
@bboure I can follow up on this internally in AWS. Will sync up with you offline.

roznalex:
> @roznalex A little follow up on this.
> Sorry for not merging this earlier. I really want to merge this asap, but I am a bit afraid that this could break again in the future.
> I opened a question on the [AWS forum](https://forums.aws.amazon.com/thread.jspa?threadID=301588). I'd like to have feedback from them before doing anything.

Hi @bboure. Are there any updates? Could we merge this PR already?

bboure:
Sorry for late response @roznalex 
AWS confirmed that it should be an Arn and they will fix the doc.

PR #238, opened 2019-04-10
Add authorizationConfig to HTTP datasource

yai333:
# Description
It is now possible to provide AuthorizationConfig as part of the HTTP data source configuration in cloud formation from recent release of AWS.
## Reference
https://docs.aws.amazon.com/appsync/latest/APIReference/API_AuthorizationConfig.html

This pull request adds AuthorizationConfig to HTTP data source configuration. 

Example of configuration as below:
```
      - type: HTTP
        name: AWSSES
        description: "send email"
        config:
          endpoint: "https://email.us-east-1.amazonaws.com/"
          authorizationConfig:
            authorizationType: "AWS_IAM"
            awsIamConfig:
              signingRegion: "us-east-1"
              signingServiceName: "ses"
          serviceRoleArn: { Fn::GetAtt: [AppSyncSESserviceRole, Arn] }
          
```

# Type of change
- New feature


bboure:
Thank you @yai333 for bringing this!
Code LGTM. Could you please add tests for this?

yai333:
Added test as per requested

index.js
```diff
@@ -643,8 +643,30 @@ class ServerlessAppsyncPlugin {
...
+        const authorizationConfig = {
+          ...(ds.config.authorizationConfig && {
+            AuthorizationConfig: {
+              ...(ds.config.authorizationConfig.authorizationType && {
+                AuthorizationType: ds.config.authorizationConfig.authorizationType,
+              }),
+              ...(ds.config.authorizationConfig.awsIamConfig && {
+                AwsIamConfig: {
+                  ...(ds.config.authorizationConfig.awsIamConfig.signingRegion && {
+                    SigningRegion: ds.config.authorizationConfig.awsIamConfig.signingRegion,
```
bboure: @yai333 Can we default the signing region to the stack's region? with fallback to account region? 
a bit like here: https://github.com/sid88in/serverless-appsync-plugin/blob/master/index.js#L287

yai333: updated as per request, please check

razor-x:
Hi, I would very much like to use this feature so I was wondering if a new minor release is planned to be released soon. Thanks

yai333:
it was merged already, you can use it without any problem

razor-x:
@yai333 Thanks for the updates, but I realize it was merged. I am asking when the next package release with this will be so I may pull it from npm.

PR #247, opened 2019-06-04
Allow AppSync tags to be set in serverless.yml in the same way that tags can be specified for other serverless resources.

hypexr:


bboure:
Thank you @hypexr 
LGTM

hypexr:
Can this be merged?  Thanks

PR #252, opened 2019-07-01
Substitutions per resolver

### LINKED ISSUE #208: Substitutions per resolver
It would be a really nice feature if we could set substitutions for each resolver separately.

**My use case**
I'm working with DynamoDB and I want to create my resolvers as generic as possible. For example when I fetch user from database I would like to fetch also country that they are from. In my database user looks like this:
```
{ 
  partitionKey: 'user_1',
  sortKey: 'user_1',
  id: 1, 
  name: 'John', 
  countryId: 1 
}
```
My country looks like this:
```
{ 
  partitionKey: 'country_1',
  sortKey: 'country_1',
  id: 1, 
  name: 'Slovenia' 
}
```
My GQL schema looks like this:
```
type User {
  id: String
  name: String
  country: Country
}

type Country {
  id: String
  name: String
}
```
To fetch country of a my user I need to set resolver for `User.country` field. Because I can not access GQL field name that is requested in query it would be really useful if we could set `substitutions` per resolver so we would be able to use one resolver for each such occasion that would look something like this:
```
#set($relationName = "$substitution") ## this would be in this example 'country'
#set($id = "Id")
#set($field = "$relationName$id")
#set($realtionId = "${ctx.source[$field]}")
#set($concat = "_")

{
    "version" : "2017-02-28",
    "operation" : "GetItem",
    "key" : {
    	"partitionKey": { "S": "$relationName$concat$realtionId" },
        "sortKey": { "S": "$relationName$concat$realtionId" },
    }
}
```

deankostomaj:
Resolves #208


deankostomaj:
@bboure is there any chance this could be merged?

deankostomaj:
@sid88in can this be merged?

index.js
```diff
@@ -872,25 +872,27 @@ class ServerlessAppsyncPlugin {
...
-    const substitutions = configVariables
+    const subs = configVariables
       .filter(value => templateVariables.indexOf(value) > -1)
       .filter((value, index, array) => array.indexOf(value) === index)
       .reduce(
         (accum, value) => Object.assign(accum, { [value]: config.substitutions[value] }),
         {},
       );
 
+    const substitutions = Object.assign({}, subs, tplSubstitutions);
```
bboure: @deankostomaj On line 879, you assign `configVariables` with a concat of `config.substitutions` and `tplConfigVariables `.
It then goes through the filters on line 887.
But in `reduce` function, you only assign values from `config.substitutions`.
These are the values that get into `subs`.
But then again you concat `subs` with `tplSubstitutions`.

This looks a bit weird. 

I think we can improve this by doing the following: 
- create a new object (e.g.: `allSubstitutions`) with that concats `config.substitutions` and `tplSubstitutions`
- assign `configVariables` with the keys of `allSubstitutions`.
- on line 890, in the `reduce` function, change `config.substitutions[value]` with `allSubstitutions[value]`.

The rest should not need to change.

deankostomaj: hey. I will fix it in the second half of December when I will have a bit more time. Sorry for the wait.

bboure:
@deankostomaj How are we with this?
Do you want some help?

I think we might make a release soon, it would be nice to include this.

deankostomaj:
@bboure is this what you had in mind? I hope I resolved the issue.

bboure:
Hi @deadcoder0904  and Happy New year!

yes, this looks to what I had in mind. 
But line 895 looks useless and `sub` should become `substitutions`.
Apart from that and that this PR needs rebasing. LGTM!

deadcoder0904:
@bboure tagged the wrong person I suppose. Happy New Year to you too :)

bboure:
ups sorry @deadcoder0904 I meant @deankostomaj of course.

deankostomaj:
@bboure I this that now everything is the way it should be. There is only one problem with lint. I don't have an idea of how to get around this.

bboure:
@deankostomaj I just added 
`/* eslint-disable no-template-curly-in-string */` at the top of the file to disable this warning.

bboure:
tested and LGTM
Thank you again @deankostomaj  and sorry for taking so long to get this merged

deankostomaj:
@bboure no problem and thank you too

PR #255, opened 2019-07-04
Exposes AppSync API ID as CF Stack Output

### LINKED ISSUE #254: Expose `ApiId` to CloudFormation Outputs
I would like to get the `ApiId` field as an output from the resulting cloudformation stack, this way I can integrate it externally with CloudFormation `describe-stack` and connect to amplify init.

lhr0909:
Resolves #254 

Tested on a full deployment and worked.

image


bboure:
LGTM 
Thank you @lhr0909

PR #294, opened 2019-12-23
Allow In-line Strings for Request Mapping Template

jzeron:
This PR is to ad inline strings for request mapping templates as an option in addition to template paths.  The string must begin with "VTL:"

### DIFF
--- a/index.js
+++ b/index.js
@@ -807,8 +807,8 @@ class ServerlessAppsyncPlugin {
         functionConfigLocation,
         tpl.response || `${tpl.type}.${tpl.field}.response.vtl`,
       );
-      const requestTemplate = reqTempl fs.readFileSync(reqTempl, 'utf8');
-      const responseTemplate = fs.readFileSync(respTempl, 'utf8');
+      const requestTemplate = this.getRequestMappingTemplate(config, tpl);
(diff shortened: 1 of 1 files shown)


index.js
```diff
@@ -836,14 +836,52 @@ class ServerlessAppsyncPlugin {
...
+        let responseMappingTemplate = fs.readFileSync(respTemplPath, 'utf8');
+        return responseMappingTemplate
+      }
+    }
+    else { // return template at default path: /mapping-templates/`${type}`/`${field}`/response.vtl
+      let respTemplPath = path.join(config.mappingTemplatesLocation, `${type}.${field}.response.vtl`);
+      let responseMappingTemplate = fs.readFileSync(respTemplPath, 'utf8');
+      return responseMappingTemplate;
+    }
+  }
```
bboure: @jzeron Can we make these two methods `DRY`er?

jzeron: Open to suggestions to make lighter...  We are on holiday this week, but can revisit soon to address unit tests & docs...

bboure:
Hey @jzeron any progress on this?
Do you want me to take over?

bboure:
This got a bit old and outdated. Will close this for now
@jzeron let me know if that's something you'd like to follow up

PR #296, opened 2020-01-10
Improved pipeline handling

markvp:
Use name for function configurations rather than type / field
Use before/after for pipeline mapping templates rather than request / response
Improve documentation of pipeline mapping templates and function configurations

### DIFF
--- a/README.md
+++ b/README.md
@@ -128,8 +128,8 @@ custom:
         # kind: UNIT (default, not required) or PIPELINE (required for pipeline resolvers)
         functions: # array of functions if kind === 'PIPELINE'
           - # function name
-        request: # request mapping template name | defaults to {field}.{type}.{pipeline ? before : request}.vtl
-        response: # response mapping template name | defaults to {field}.{type}.{pipeline ? after : response}.vtl
+        request: # request mapping template name | defaults to {type}.{field}.request.vtl
+        response: # response mapping template name | defaults to {type}.{field}.response.vtl
         # When caching is enaled with `PER_RESOLVER_CACHING`,
         # the caching options of the resolver.
         # Disabled by default.
@@ -284,8 +284,8 @@ custom:
     mappingTemplates:
       - type: Query
         field: testPipelineQuery
-        request: './mapping-templates/before.vtl' # the pipeline's "before" mapping template
-        response: './mapping-templates/after.vtl' # the pipeline's "after" mapping template
+        request: './mapping-templates/before.vtl' # the pipeline's "before" mapping template, defaults to {type}.{field).request.vtl
+        response: './mapping-templates/after.vtl' # the pipeline's "after" mapping template, defaults to {type}.{field}.response.vtl
         kind: PIPELINE
         functions:
           - authorizeFunction
@@ -293,12 +293,12 @@ custom:
     functionConfigurations:
       - dataSource: graphqlLambda
         name: 'authorizeFunction'
-        request: './mapping-templates/authorize-request.vtl'
-        response: './mapping-templates/common-response.vtl'
+        request: './mapping-templates/authorize-request.vtl' # defaults to {name}.request.vtl
+        response: './mapping-templates/common-response.vtl' # defaults to {name}.response.vtl
       - dataSource: dataTable
         name: 'fetchDataFunction'
-        request: './mapping-templates/fetchData.vtl'
-        response: './mapping-templates/common-response.vtl'
+        request: './mapping-templates/fetchData.vtl' # defaults to {name}.request.vtl
+        response: './mapping-templates/common-response.vtl' # defaults to {name}.response.vtl
 ```
 
 ## ▶️ Usage
--- a/README.md
+++ b/README.md
@@ -125,8 +125,7 @@ custom:
       - dataSource: # data source name
         type: # type name in schema (e.g. Query, Mutation, Subscription)
         field: getUserInfo
-        # pipeline resolvers use kind: PIPELINE, otherwise kind is not required
-        # kind: pipeline
+        # kind: UNIT (default, not required) or PIPELINE (required for pipeline resolvers)
         functions: # array of functions if kind === 'PIPELINE'
           - # function name
         request: # request mapping template name | defaults to {field}.{type}.{pipeline ? before : request}.vtl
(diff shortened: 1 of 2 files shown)


src/index.js
```diff
@@ -825,11 +825,11 @@ class ServerlessAppsyncPlugin {
     return flattenedFunctionConfigurationResources.reduce((acc, tpl) => {
       const reqTemplPath = path.join(
         functionConfigLocation,
-        tpl.request || `${tpl.type}.${tpl.field}.request.vtl`,
+        tpl.request || `${tpl.name}.request.vtl`,
       );
       const respTemplPath = path.join(
         functionConfigLocation,
-        tpl.response || `${tpl.type}.${tpl.field}.response.vtl`,
+        tpl.response || `${tpl.name}.response.vtl`,
```
bboure: @markvp I understand you point.
Thinking about this now, I guess that the previous implementation did not really make sense
However, this should still be considered a breaking change.
Can we add backward compatibility?

src/index.js
```diff
@@ -872,8 +872,10 @@ class ServerlessAppsyncPlugin {
     const flattenedMappingTemplates = config.mappingTemplates
       .reduce((accumulator, currentValue) => accumulator.concat(currentValue), []);
     return flattenedMappingTemplates.reduce((acc, tpl) => {
-      const reqTemplPath = path.join(config.mappingTemplatesLocation, tpl.request || `${tpl.type}.${tpl.field}.request.vtl`);
-      const respTemplPath = path.join(config.mappingTemplatesLocation, tpl.response || `${tpl.type}.${tpl.field}.response.vtl`);
+      const reqSuffix = tpl.kind === 'PIPELINE' ? 'before' : 'request';
+      const respSuffix = tpl.kind === 'PIPELINE' ? 'after' : 'response';
+      const reqTemplPath = path.join(config.mappingTemplatesLocation, tpl.request || `${tpl.type}.${tpl.field}.${reqSuffix}.vtl`);
+      const respTemplPath = path.join(config.mappingTemplatesLocation, tpl.response || `${tpl.type}.${tpl.field}.${respSuffix}.vtl`);
```
bboure: @markvp This however **is** a breaking change.
People probably already rely on the fact that their `before`/`after` mapping templates are called `request`/`response`.

I see no problem in adding support for `.before`/`.after` suffixes but we must keep backward compatibility with `request`/`response`.

README.md
```diff
@@ -123,10 +123,12 @@ custom:
     mappingTemplatesLocation: # defaults to mapping-templates
     mappingTemplates:
       - dataSource: # data source name
-        type: # type name in schema (e.g. Query, Mutation, Subscription)
+        type: # type name in schema (e.g. Query, Mutation, Subscription, PIPELINE)
```
bboure: @markvp `type` does not take `PIPELINE` as a value, it's under `kind`.

markvp: oops.  fixed.

src/index.js
markvp: Given that functions don't have a type or a field, I'm not sure how anyone could have used this functionality in the past? I certainly couldn't get it to work.
Given that, I don't really see this is a breaking change as much as it is a fix of unusable code (no offence!)

src/index.js
markvp: Hmm. I understand the desire for backwards compatibility. Not sure the best way to implement that. Perhaps checking for file existence first?

src/index.js
bboure: @markvp After giving it a second thought and having a deeper analysis on this, I think you are right. 
I got that those fields are indeed not present in pipeline definitions, but I thought that someone might have used them anyway (as a hack) in order to auto-generate the filename. But I realize that it would make no sense, since you could just place the filename directly instead.

I'll accept this change 👍

README.md
```diff
@@ -125,8 +125,12 @@ custom:
       - dataSource: # data source name
         type: # type name in schema (e.g. Query, Mutation, Subscription)
         field: getUserInfo
-        request: # request mapping template name
-        response: # response mapping template name
+        # pipeline resolvers use kind: PIPELINE, otherwise kind is not required
+        # kind: pipeline
```
bboure: @markvp not required when UNIT, but still it would be accepted. I guess it does not hurt to specify it. 
Also, is ahs to be `PIPELINE` or `UNIT` (uppercase)

otherwise, [it will fail](https://github.com/sid88in/serverless-appsync-plugin/blob/master/src/index.js#L909)

src/index.js
bboure: @markvp 🤔 Well, in CloudFormations (and in this plugin), even if in the doc they refer to it as `before` and `after`, the fields that hold these template paths are still called `request` and `response`.
So I think the right thing to do is to avoid complexity (and confusion) and stick to that.

markvp: ok. removed before/after update for now

README.md
markvp: fixed

bboure:
Thank you @markvp  LGTM

Yojimb0:
Hi @bboure, any ETA for when a version bump will include this change ? It appears that v1.2.0 still expects pipeline functions to use `${tpl.type}.${tpl.field}.request.vtl` as default naming convention (instead of upcoming `${tpl.name}.response.vtl`)

Cheers

bboure:
@Yojimb0 There is a bunch of PR pending merging. I'll try to find some time to merge them and version bump this week.

Yojimb0:
Sweet, thanks :)

bboure:
Released in 1.3.0 🎉

morficus:
I recently updated to v1.3.0 and this broke my ability to deploy :-(
After way too much time of debugging and searching... I came across this PR and realize that it introduced a **BREAKING CHANGE** which is not documented anywhere

> Use name for function configurations rather than type / field

All of my functions are currently named using snake-case (ie: `list-cars`, `find-car-by-id`, `delete-car`). I had to rename all 57 of my function to camelCase (`listCars`, `findCarById`, etc.)

Prior to v1.3.0... that was not a problem because this plugin used `type`+`fieldName` to generate the deployed function name, which always resulted in something AppSync considered valid. But now that it uses the `name` field directly... all 57 of my functions are now considered to have "invalid names".

While I agree that using `name` over `type`+`fieldName` does make more sense semantically... it would have been really great to either (1) put a big far warning about upgrading or (2) automatically convert names in to AppSync-safe names.

I see there is a PR to fix this (PR #325), but I wanted to express my frustration in a way that would (hopefully) help others that come across this issue thread via google.

PR #306, opened 2020-02-23
Updating README

morficus:
AppSync does not support a value of "NONE" for the authentication type.
Trying to use the value present in the README will result in a validation error: "appSync property `authenticationType` is missing or invalid."


bboure:
Thank you @morficus
You are right.
Can we list all possible values thought?

Issue #321, opened 2020-04-07
Function names prefixed with "GraphQlFunctionConfiguration"


tsmith-rv:
All of my Appsync function names are prefixed with `GraphQlFunctionConfiguration` once they're created in AWS. From the code in the plugin, it looks like this is intended behavior. Is there a specific reason for this? I wouldn't mind opening a PR to remove the prefix if everyone was open to that, but my assumption is that this exists for a reason, or maybe you can create the function without a prefixed name and I'm just missing it.

My YML
```yml
- name: GetActiveLinksByLocation
  dataSource: GetActiveLinksByLocation
```

Expected function name in AWS: `GetActiveLinksByLocation`
Actual function name in AWS: `GraphQlFunctionConfigurationGetActiveLinksByLocation`

bboure:
Hi @tsmith-rv  Thank you for bringing this to our attention.

All resource names come with a prefix in order to avoid clashes in names. 
It looks like [here](https://github.com/sid88in/serverless-appsync-plugin/blob/master/src/index.js#L875) we are using the Resource Name as the function name. It looks like a mistake to me, and I guess it could be removed.

I just wonder what impact this could have on backward compatibility. I don't feel like it should be an issue.

We would be happy to accept a PR that removes it if you can test and confirm that it does not break existing stacks.

tsmith-rv:
> Hi @tsmith-rv Thank you for bringing this to our attention.
> 
> All resource names come with a prefix in order to avoid clashes in names.
> It looks like [here](https://github.com/sid88in/serverless-appsync-plugin/blob/master/src/index.js#L875) we are using the Resource Name as the function name. It looks like a mistake to me, and I guess it could be removed.
> 
> I just wonder what impact this could have on backward compatibility. I don't feel like it should be an issue.
> 
> We would be happy to accept a PR that removes it if you can test and confirm that it does not break existing stacks.

On it!

tsmith-rv:
I'm about to open the PR now. It works well. It basically just seems to treat it as a normal function renaming operation. Normally when you rename an Appsync function, as the stack updates, you temporarily have two functions (the original and a new one with the updated name) and as long as everything is successful with the stack update, it switches to the new one and deletes the old one. 

Like this (before I updated the code):

Original Function
[...]

PR #323, opened 2020-04-11
Update Created Function Naming

tsmith-rv:
fixes https://github.com/sid88in/serverless-appsync-plugin/issues/321 — I updated it so that when Appsync functions are created it uses the name specified in the yml file in the `name` field, as opposed to before where it was using the `name` field prefixed with `GraphQlFunctionConfiguration`, which is the resource name.


jaredready:
@sid88in when can we expect another release of this plugin? I'd love to be able to clean up my function names :)

bboure:
@jaredready We'll try to release a new version soon.
Thank yo ufor you patience

bboure:
Released in 1.3.0 🎉

jaredready:
Thanks @bboure !

PR #325, opened 2020-04-21
Fix for v1.3.0 breaking change: Sanitize function name

mfogel:
It appears #321 broke deployment for projects that are using characters like `.` or `-` in their serverless.yml for function configuration names.

There's already a sanitization function present and in use in this project, this PR just applies it to the function name.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -875,7 +875,7 @@ class ServerlessAppsyncPlugin {
           Type: 'AWS::AppSync::FunctionConfiguration',
           Properties: {
             ApiId: { 'Fn::GetAtt': [logicalIdGraphQLApi, 'ApiId'] },
-            Name: tpl.name,
+            Name: this.getCfnName(tpl.name),
             DataSourceName: { 'Fn::GetAtt': [logicalIdDataSource, 'Name'] },
             RequestMappingTemplate: this.processTemplate(
               requestTemplate,


bboure:
Thank you @mfogel 
Is there any reason why your functions have special characters in them?
I would say this is rather unexpected.
Now, I agree that #321 was a breaking change in this case, as before it was sanitized (even if this was not meant to).

I wonder however if using `getCfnName()` is the right choice.

According to what I see in AWS console, the regex for `name` is `[_A-Za-z][_0-9A-Za-z]*`

I'll try to have a deeper look at this tomorrow

mfogel:
Yeah, it looks like `getCfnName()` does a bit more sanitization than it needs to as it removes underscores, which AWS's regex allows.

I imagine `getCfnName()` could be relaxed to also allow for underscores, though I would view that a separate task than this one. `getCfnName()` is used also used to sanitize other fields other than this one, and I'm not 100% sure that AWS applies the same regex to all fields. Probably does, but I don't know that.

mfogel:
@bboure any reason in particular not to accept this PR? Is there a downside to using `getCfnName()` to sanitize the function names, as was done before the merge of #321 ?

bboure:
Sorry @mfogel I kinda forgot about this. 
I understand that 1.3.0 was a breaking change for you. Although, I am not sure if we should transform names like this. This could be unexpected. I see 1.3.0 rather a fix.
I am also afraid that we would break it again for other users by changing it.

What would be ideal for me would be to:
- keep the name as is (from the serverless.yml) file
- if the name is incorrect (invalid characters), throw an error.

This would force the serverless file to have valid names instead of "forcing" a valid name. This is probably what we should have done from the beginning.

Would that be an issue for you to rename your functions in your stack instead of "fixing" it in the puggin?  

PS: sorry for closing the PR, I misclicked a button 🤦

mfogel:
Hi @bboure, totally valid concerns and it is not hard for me to rename my function names. A few things to keep in mind:

- I don't think you need to worry about renaming everyone's functions. This is exactly what #321 did, which caused this breaking change. #321 did not break the plugin for most users
- But in any case, accepting this PR does _not_ actually rename everyone's functions. It only renames them for people who have `.` or `-` in them or other characters that `getCfnName()` removes
- why do we have the function `getCfnName()` if not for this purpose? It was added in 27668453, with this exact purpose... as I understand. To "Sanitize non cfn friendly resource names"

If we want to change the behavior of this plugin to stop sanitizing resource names - ok, fine. Then...

- shouldn't we stop using `getCfnName()` completely, for consistency? It's used to sanitize resource names in ~7 other places. It seems kinda odd to just stop using it in just this one place
- should we release a new major version of the plugin? This is a backward incompatibility.

Cheers, appreciate your help.

mfogel:
BTW, this sounds totally reasonable future behavior of the plugin from my perspective:

> What would be ideal for me would be to:
>
> - keep the name as is (from the serverless.yml) file
> - if the name is incorrect (invalid characters), throw an error.

It's just that:

- this change should applied everywhere, not just to this one instance of a resource name
- this change should be part of a new major version release (ie: `2.0`) because this is a backward incompatibility

My recommendation would be to fix this issue for the `1.X` version of the plugin, then set up the above change as part of a `2.0` release.

bboure:
> why do we have the function getCfnName() if not for this purpose? It was added in 2766845, with this exact purpose... as I understand. To "Sanitize non cfn friendly resource names"

It is meant to generate Resource **IDS** for cloudformation based on their names (hence the name: `getCfnName()`). Resources **names** should not be changed and if that happens, it is certainly a mistake (precisely that was the  reason of #323)

> should we release a new major version of the plugin? This is a backward incompatibility.

I have been considering a major release for some time now. There are many things that are kept for backward compatibility reasons only and make the code harder to maintain and understand and probably a bunch of things we could improve/change and would break BC.

Right now this is not a priority though.

> But in any case, accepting this PR does not actually rename everyone's functions. It only renames them for people who have . or - in them or other characters that getCfnName() removes 

Totally true. So I guess there would be no harm in accepting this PR.

morficus:
I, for one, would love for this PR to be accepted 🙏 
especially since transparently addresses a breaking change introduced in a minor release.

src/index.js
```diff
@@ -875,7 +875,7 @@ class ServerlessAppsyncPlugin {
           Type: 'AWS::AppSync::FunctionConfiguration',
           Properties: {
             ApiId: { 'Fn::GetAtt': [logicalIdGraphQLApi, 'ApiId'] },
-            Name: tpl.name,
+            Name: this.getCfnName(tpl.name),
```
morficus, line 878: I saw in the discuss thread that there was some concerns around using `getCfnName`.
Alternatively, sanitization could be done using [`_.camelCase `](https://lodash.com/docs/4.17.15#camelCase) or [`_.snakeCase`](https://lodash.com/docs/4.17.15#snakeCase)

bboure:
@mfogel @morficus  Sorry I did not have time to handle this recently.
I understand this is still an issue, and for more people. 
I'll just accept this change as-is for now. 

Thanks to both of you.

bboure:
Thank you

PR #326, opened 2020-04-29
Update get-stack-value.js to fetch the stackname from provider.naming

moelholm:
This change ensures that the stack name is fetched from serverless' provider.naming object.

It fixes a problem where the playground cannot be started against stacks with "custom" stack names (as opposed to being automatically named by serverless).


bboure:
Thanks @moelholm 
LGTM

bboure:
Thank you

PR #332, opened 2020-05-21
Update readme for dataSource type

veloware:
Elaborating on how to attach a resolver to a single field on a custom type.  Showing that not only the standard Query, Mutation and Subscription types can be used.


bboure:
@veloware 
Thanks for this. Makes the doc clearer.

bboure:
Thank you

PR #339, opened 2020-07-13
Documents functionConfigurationsLocation

thomasmichaelwallace:
`functionConfigurationsLocation` is a useful addition (especially when organizing larger projects).

I thought I'd add it to the documentation so that people can discover it more easily.


bboure (approved):
Sorry. I missed that @thomasmichaelwallace

Issue #317, opened 2020-03-25
Triple quote string literal descriptions cause late validation failure


thomasmichaelwallace:
First- thanks for all the work on this ace project!

AppSync's support for the GraphQL SDL is a little out of date. Specially it doesn't support string literals (triple quote) as descriptions: https://github.com/aws/aws-appsync-community/issues/38

If you use this [now best practace](https://spec.graphql.org/June2018/#example-64e5a) approach with this plug-in the schema validates (because, well, it is valid!) but your deployment will fail with:

```
ServerlessError: An error occurred: GraphQlSchema - Schema Creation Status is FAILED with details: Failed to parse schema document - ensure it's a valid SDL-formatted document..
```

Luckily I was aware of this problem, so I didn't lose too much time. But I think it would be nice if this plugin either:

1. Automatically stripped triple quote string literals before deploying the stitched schema [preferred]
2. Failed validation if the schema includes triple quote string literals.

thomasmichaelwallace:
As part of my own deployment I'm solving the problem by pre-processing (removing the offending descriptions).

I'm more than happy to fold this into a PR if you agree it's something that should be part of this plugin.

thomasmichaelwallace:
Actually - I think you might already be doing this, and I might have another problem.

Let me report back.

thomasmichaelwallace:
For reference, this is what I needed to do get my schema to pass appsync's internal definition:

```js
  const conformSchema =
    (schema) => schema
      .replace(/"""[^"]*"""\n/g, '') // appsync does not support the new style descriptions
      .replace(/#.*\n/g, '') // appysnc does not support old-style # comments in enums, so remove them all
      .replace(/ *& */g, ', ') // appsync does not support the standard '&', but the "unofficial" ',' join for interfaces:
```

bboure:
Fixed in #340

PR #340, opened 2020-07-13
Fix: Triple quote string literal descriptions cause late validation failure

thomasmichaelwallace:
The GraphQL validation function, etc. is all based on GraphQL v14.5.8. Unfortunately there are some key differences between this schema version are the (now quite old) version AppSync uses. This results in cases where the validation step passes, and then the stack update fails.

This patch specifically addresses the three cases I have found (#317):
 * AppSync does not support the new style (`""" triple string literal """`) documentation/descriptions
 * AppSync does not support old-style (`#`) comments in Enums
 * AppSync uses the unofficial comma (install of the standard '&') join for interfaces

I've been using this patch in against our production schema for quite some time, but let me know if you think it's worth integrating.

(I appreciate there's now a GraphQL v15, but that's a whole other story!)

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -838,7 +838,7 @@ class ServerlessAppsyncPlugin {
     const appSyncSafeSchema = config.schema
       .replace(/"""[^"]*"""\n/g, '') // appsync does not support the new style descriptions
       .replace(/#.*\n/g, '') // appysnc does not support old-style # comments in enums, so remove them all
-      .replace(/ *& */g, ', ') // appsync does not support the standard '&', but the "unofficial" ',' join for interfaces
+      .replace(/ *& */g, ', '); // appsync does not support the standard '&', but the "unofficial" ',' join for interfaces
     return {
       [logicalIdGraphQLSchema]: {
         Type: 'AWS::AppSync::GraphQLSchema',
--- a/src/index.js
+++ b/src/index.js
@@ -835,11 +835,15 @@ class ServerlessAppsyncPlugin {
   getGraphQLSchemaResource(config) {
     const logicalIdGraphQLApi = this.getLogicalId(config, RESOURCE_API);
     const logicalIdGraphQLSchema = this.getLogicalId(config, RESOURCE_SCHEMA);
+    const appSyncSafeSchema = config.schema
+      .replace(/"""[^"]*"""\n/g, '') // appsync does not support the new style descriptions
+      .replace(/#.*\n/g, '') // appysnc does not support old-style # comments in enums, so remove them all
+      .replace(/ *& */g, ', ') // appsync does not support the standard '&', but the "unofficial" ',' join for interfaces
     return {
       [logicalIdGraphQLSchema]: {
         Type: 'AWS::AppSync::GraphQLSchema',
         Properties: {
-          Definition: config.schema,
+          Definition: appSyncSafeSchema,
           ApiId: { 'Fn::GetAtt': [logicalIdGraphQLApi, 'ApiId'] },
         },
       },
--- a/__snapshots__/index.test.js.snap
+++ b/__snapshots__/index.test.js.snap
@@ -403,6 +403,29 @@ Object {
 }
 `;
 
+exports[`appsync config Schema is transformed into App Sync compatible syntax 1`] = `
+Object {
+  "GraphQlSchema": Object {
+    "Properties": Object {
+      "ApiId": Object {
+        "Fn::GetAtt": Array [
+          "GraphQlApi",
+          "ApiId",
+        ],
+      },
+      "Definition": "
+                    type Thing implements One, Another {
+            hello: ID!
+          }
+                    enum Method {
+            DELETE             GET           }
+        ",
+    },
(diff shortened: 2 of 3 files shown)


thomasmichaelwallace:
Oops; I thought I had linting on, but I obviously didn't. I'll get semi-colon fixed and update the PR.

src/index.js
```diff
@@ -835,11 +835,15 @@ class ServerlessAppsyncPlugin {
   getGraphQLSchemaResource(config) {
     const logicalIdGraphQLApi = this.getLogicalId(config, RESOURCE_API);
     const logicalIdGraphQLSchema = this.getLogicalId(config, RESOURCE_SCHEMA);
+    const appSyncSafeSchema = config.schema
+      .replace(/"""[^"]*"""\n/g, '') // appsync does not support the new style descriptions
+      .replace(/#.*\n/g, '') // appysnc does not support old-style # comments in enums, so remove them all
+      .replace(/ *& */g, ', '); // appsync does not support the standard '&', but the "unofficial" ',' join for interfaces
```
bboure, line 841: @thomasmichaelwallace what do you think if we throw an error at the plugin level instead of automagically fixing the schema?
Using `.replace()` looks both hacky and dangerous. 

Best practice would probably be to use an appsync-safe schema.

I'm just thinking out loud here. Let me know what you think based on your experience.

bboure:
Thank you @thomasmichaelwallace for this contrib. Sorry, I must have missed the original issue.
There are some known issues indeed with the graphql version that appsync is using. 
I already [opened an issue](https://github.com/aws/aws-appsync-community/issues/71) about that.

We should probably push to get an upgraded version.

thomasmichaelwallace:
> Thank you @thomasmichaelwallace for this contrib. Sorry, I must have missed the original issue.
> There are some known issues indeed with the graphql version that appsync is using.
> I already [opened an issue](https://github.com/aws/aws-appsync-community/issues/71) about that.
> 
> We should probably push to get an upgraded version.

I guess I'm not really holding my breath on that one, given there's also https://github.com/aws/aws-appsync-community/issues/38 - which is over a year old today :)

src/index.js
thomasmichaelwallace: I guess throwing an error at plugin level if triple quotes, or interface joins with '&'s, etc. addresses the late validation issue (i.e. that the plugin validator lets through app-sync invalid schemas).

I can see why you might not want the replaces in there (I haven't found a library that translates June 2018 spec schemas back down to whatever it is that AWS implemented). That we've been using it for about half a year in production without any issues is probably a bit too much of a "works on my machine ™️" defense.

thomasmichaelwallace:
Re: support for people using June 2018 tooling, etc. Are you open to providing some hooks for people like me who want hacky `.replace`'s so they can have the best of both worlds?

bboure:
> That we've been using it for about half a year in production without any issues is probably a bit too much of a "works on my machine ™️" defense.

Yeah, in fact I am just afraid that we would replace things that should not. Though, the risks seem limited here. 

> What catches me out is that a lot of the tooling wants the June 2018 spec.

Yeah, you got a point there. I am not sure how to "fix" that though. 

Quick question:
If you use the AppSync specs in your schema directly, does the validation pass? or does the June 2018 validator fails?

thomasmichaelwallace:
> Quick question:
> If you use the AppSync specs in your schema directly, does the validation pass? or does the June 2018 validator fails?

The June 2018 validator fails when using the unofficial "," implementation syntax (from what I can tell, this was _never_ the spec), for example:

```graphql
# an example app-sync valid schema that won't validate:
type Query {
    # get a node by ide
  versionedNode(
    id: ID! # NodeId to access
  ): AVersionedNode
}

# something that is versioned
interface Versioned {
  # the version
  version: Int!
}

# something that is accessible by id
interface Node {
  # the id
  id: ID!
}

# implements interfaces using June 2018 unsupported comma notation
type AVersionedNode implements Node, Versioned {
  type: ID!
  version Int!
  name: String
}
```

```sh
serverless validate-schema
AppSync Plugin: Syntax Error: Unexpected Name "Versioned"

GraphQL request:22:38
21 | # implements interfaces using unsupported comma
22 | type AVersionedNode implements Node, Versioned {
   |                                      ^
23 |   type: ID!
```

thomasmichaelwallace:
> Yeah, in fact I am just afraid that we would replace things that should not. Though, the risks seem limited here.

While walking the dog, it did occur to me that the transformation could be optional (via a configuration flag).

bboure:
> The June 2018 validator fails when using the unofficial "," implementation syntax (from what I can tell, this was never the spec),

I was afraid of that, :(

> While walking the dog, it did occur to me that the transformation could be optional (via a configuration flag).
I though of that too, but it is an issue if the validator fails on the unofficial specs. 

I'll try to think of a workaround, otherwise, we'll have to hack it your way 🗡️ 
thanks for the details and the investigation

bboure:
@thomasmichaelwallace Il''merge this as-is for now. 
I guess this is better than nothing.
Thanks you.

bboure:
Published in v1.4.0

PR #350, opened 2020-08-07
Feat/direct lambdas

### LINKED ISSUE #348: FEATURE REQUEST : Support for Direct Lambda Resolvers
VTL Request/Reponse Mapping is now optional.
Given the convenience feature of defaults for request response vtl, 
there probably needs to be a update to the plugin to support direct lambda resolvers, correct?

[https://aws.amazon.com/blogs/mobile/appsync-direct-lambda/](https://aws.amazon.com/blogs/mobile/appsync-direct-lambda/)

bboure:
Fixes #348 

Cc: @kziemski


osjupiter:
I tried using this branch feature , but it didn't work.
Maybe the property name `defaultMappingTemplates` in getConfig is wrong ( shoude be `defaultTemplates`)

bboure:
Good catch @osjupiter  I will fix this.

bboure:
This should work now.
I just tested both default templates and direct lambda features. Seems to work fine.

bboure:
Published in v1.4.0

iamdavidmartin:
Can you share a snippet of a serverless.yml with this working?

bboure:
@iamdavidmartin You can just set false as the request/response of your mappingTemplate:

```yml
    mappingTemplates:
      - dataSource: myLambdaFunction
        type: Query
        field: myField
        request: false
        response: false
```

For a full example of a serverless.yml file, you can check the [example](https://github.com/sid88in/serverless-appsync-plugin/blob/master/example/serverless.yml)

iamdavidmartin:
Thanks for the reply. I'm still having some trouble. I can't seem to get the dataSource to find my lambda function. 

Here is my mapping:

```
      - type: Mutation
        dataSource: removeImageFromS3Function
        field: removeImageFromS3
        request: false
        response: false
```


And this is the error I get. Notice that "removeImageFromS3Function" is prepended with "GraphQlDs". I've tried renaming my function from "removeImageFromS3Function" to "GraphQlDsremoveImageFromS3Function.

`  An error occurred: AppSyncNestedStack - Template error: instance of Fn::GetAtt references undefined resource GraphQlDsremoveImageFromS3Function.`

I'm really flummoxed. I'm using split stacks and the mutation and the function are defined in separate json files. Could that be the issue?

bboure:
@iamdavidmartin Note that you still need to define your datasource, even with direct lambda:

```yml
functions:
  # Define the lambda function
  removeImageFromS3Function:
    handler: dist/src/functions/removeImageFromS3/index.handler
    runtime: nodejs12.x
    role: AdminRole
    timeout: 60
    layers:
      - { Ref: ModulesLambdaLayer 

dataSources:
  # Define a datasource of type 'LAMBDA' that uses the function
  - type: AWS_LAMBDA
    name: removeImageFromS3
    description: 'Lambda DataSource'
    config:
      functionName: removeImageFromS3Function

mappingTemplates:
  # use the datasource in the mappingTemplate
  - type: Mutation
    dataSource: removeImageFromS3
    field: removeImageFromS3
    request: false
    response: false
```

muzaffar622:
@bboure I did as u noted but still not works.

```
mappingTemplates:
  - dataSource: myLambda
     type: Query
     field: accounts
     request: false
     response: false


  - type: AWS_LAMBDA
        name: myLambda
        description: "lambda data source"
        config:
          functionName: graphql


functions:
  graphql:
    handler: handler.graphqlHandler
```
error 

`ENOENT: no such file or directory, open 'mapping-templates/Query.accounts.request.vtl'`

iamdavidmartin:
Ah, yes. Got it working. As usual, the problem existed between keyboard and chair.

PR #351, opened 2020-08-07
Generate ConditionCheckItem IAM role to support ConditionCheck

lulzneko:
AppSync's TransactWriteItems now support DynamoDB's ConditionCheck.

It generates the IAM Role needed to run ConditionCheck.

[Resolver Mapping Template Reference for DynamoDB - AWS AppSync](https://docs.aws.amazon.com/appsync/latest/devguide/resolver-mapping-template-reference-dynamodb.html#aws-appsync-resolver-mapping-template-reference-dynamodb-transact-write-items)

[Using IAM with DynamoDB Transactions - Amazon DynamoDB](https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/transaction-apis-iam.html)

### DIFF
--- a/__snapshots__/index.test.js.snap
+++ b/__snapshots__/index.test.js.snap
@@ -518,6 +518,7 @@ Object {
                   "dynamodb:UpdateItem",
                   "dynamodb:BatchGetItem",
                   "dynamodb:BatchWriteItem",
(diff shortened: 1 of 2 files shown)


bboure:
Published in v1.4.0

PR #362, opened 2020-10-13
Fix an important typo in README.md

anasqadrei:
fix an important typo


bboure (approved):
Ups. Thank you! @anasqadrei

PR #372, opened 2020-11-24
Support `functionAlias` in lambda data sources.

### LINKED ISSUE #365: Lambda functionName does not pick up alias
By providing `provisionedConcurrency` or `reservedConcurrency` to the function itself, serverless will create an alias for the lambda. But this plugin does not automatically pick up the generated alias/qualifiers and only connects to the $LATEST version without scaling/throttling.

Is it supposed to be able to pick it up, or is there any other ways to do it?

vicary:
Resolves #365

### DIFF
--- a/.all-contributorsrc
+++ b/.all-contributorsrc
@@ -532,6 +532,15 @@
       "contributions": [
         "code"
       ]
+    },
+    {
+      "login": "vicary",
+      "name": "Vicary A.",
+      "avatar_url": "https://avatars0.githubusercontent.com/u/85772?v=4",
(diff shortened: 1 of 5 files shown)


bboure (approved):
Thanks! @vicary  LGTM

bboure:
@vicary please fix the lining, and this is good to go.
Bonus point for adding tests.

vicary:
first time doing snapshot tests, feels quite usable tbh.

bboure:
Thanks again @vicary

Issue #357, opened 2020-09-29
Allow to select which API key for API_KEY authentication


Jackychans:
Hi everyone,

I'm building Saas app model where each client should have their own API_KEY specified expiry dateTime  as additional authorization method (AWS Cognito User Pool is used by default authorization). 

I can see we have one field API_Key but have no idea how appsync works out on selecting specifically which API_KEY corresponding for each clients.

Another scenario using API_Key as additional authorization is to allow my mobile apps accessing AppSync with different API_KEY than web apps. Each app should use different API Keys.

Any advise ? really appreciate

bboure:
Hi @Jackychans 
There are several issues here I think
1) AppSync does support several API keys definitions, and that is definitely something we should support (Currently, we don't)
2) Currently, the plugin generates an API key with a one year expiry date by default. I understand you might want to set that yourself. This is also something we should support. 
3) I imagine you might want to define yourself custom API keys for each microservice/app.
This one is currently not possible (API key is generated randomly). See [this issue](https://github.com/aws/aws-appsync-community/issues/130) for more details. 

Will look into 1) and 2) when I can, unless you want to open a PR yourself??

bboure:
@Jackychans Please check the new feature available in v1.5.0
See the [Api key Management](https://github.com/sid88in/serverless-appsync-plugin#managing-api-keys) section in the README

sid88in:
This is supported on v2. You can define multiple named API keys, each with its own expiry —
exactly your mobile-vs-web case:

```yml
appSync:
  apiKeys:
    - name: mobile
      description: Mobile app key
      expiresAfter: 30d
    - name: web
      description: Web app key
      expiresAfter: 365d
```

Reference a specific key's generated value with `${appsync:apiKey.mobile}` etc., and rotate
or revoke them independently. See `doc/API-keys.md` and the `examples/api-keys-multiple`
project (there's an e2e test for it too).

The one thing that still isn't possible is choosing the literal key *string* yourself —
AppSync generates the value and there's no API to set it (aws-appsync-community#130). And note
a key isn't scoped to particular operations: any valid key satisfies `API_KEY` auth, so which
client gets which key is a distribution concern on your side.

Closing as implemented — thanks @Jackychans.

PR #373, opened 2020-12-07
feat: add apiKeyRepairEnabled AppSync configuration option

bsantare:
Set apiKeyRepairEnabled to `true` to automatically create a new, unique AppSync API key CFN object if the plugin-maintained CFN object is associated with an AppSync API key id that no longer exists.  This can happen on API key expiration or if an API key is manually deleted from the AWS console.  If this condition is present, the deployment will fail without this option enabled because CFN will try to look up the deleted API key by ID and fail.  This option is only relevant for authenticationType === API_KEY

* Modify mergeCustomProviderResources hook to be async
* Modify addResources to allow async calls
* Modify getApiKeyResources to handle key repairs when enabled
* Modify getApiKeyOutputs to use the resolvedLogicalApiKeyId resolved in getApiKeyResources

### DIFF
--- a/README.md
+++ b/README.md
@@ -79,6 +79,12 @@ custom:
   appSync:
     name:  # defaults to api
     # apiKey # only required for update-appsync/delete-appsync
+    # Set to `true` to automatically create a new, unique AppSync API key CFN object if the plugin-maintained
+    # CFN object is associated with an AppSync API key id that no longer exists.  This can happen on API key exipration
+    # or if an API key is manually deleted from the AWS console.  If this condition is present, deployment will
+    # fail without this option enabled because CFN will try to look up the deleted API key by ID and fail.
+    # This option is only relevant for authenticationType === API_KEY
+    # apiKeyRepairEnabled: true
     authenticationType: API_KEY or AWS_IAM or AMAZON_COGNITO_USER_POOLS or OPENID_CONNECT
     schema: # schema file or array of files to merge, defaults to schema.graphql
     # Caching options. Disabled by default
@@ -159,10 +165,10 @@ custom:
         #    - lambdaFunctionArn: The Arn for the Lambda function to use as the Conflict Handler.
         #    - conflictHandler: The Conflict Resolution strategy to perform in the event of a conflict.
         sync:
-          conflictDetection: VERSION # https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-appsync-resolver-syncconfig.html 
+          conflictDetection: VERSION # https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-appsync-resolver-syncconfig.html
           conflictHandler: OPTIMISTIC_CONCURRENCY # when not using lambda conflict handler choose The Conflict Resolution strategy to perform in the event of a conflict. OPTIMISTIC_CONCURRENCY / AUTOMERGE / LAMBDA
           functionName: graphql # The function name in your serverless.yml. Ignored if lambdaFunctionArn is provided.
-          lambdaFunctionArn: "arn:aws:lambda:{REGION}:{ACCOUNT_ID}:myFunction"       
+          lambdaFunctionArn: "arn:aws:lambda:{REGION}:{ACCOUNT_ID}:myFunction"
 
       - ${file({fileLocation}.yml)} # link to a file with arrays of mapping templates
     functionConfigurationsLocation: # defaults to mappingTemplatesLocation (mapping-templates)
--- a/__snapshots__/get-config.test.js.snap
+++ b/__snapshots__/get-config.test.js.snap
@@ -6,6 +6,7 @@ Array [
     "additionalAuthenticationProviders": Array [],
     "apiId": undefined,
     "apiKey": undefined,
+    "apiKeyRepairEnabled": undefined,
     "authenticationType": "AWS_IAM",
     "caching": undefined,
     "dataSources": Array [],
@@ -100,6 +101,7 @@ Array [
     "additionalAuthenticationProviders": Array [],
     "apiId": undefined,
     "apiKey": undefined,
+    "apiKeyRepairEnabled": undefined,
     "authenticationType": "AWS_IAM",
     "caching": undefined,
     "dataSources": Array [],
@@ -225,6 +227,7 @@ Array [
     "additionalAuthenticationProviders": Array [],
     "apiId": undefined,
     "apiKey": undefined,
+    "apiKeyRepairEnabled": undefined,
     "authenticationType": "AWS_IAM",
     "caching": undefined,
     "dataSources": Array [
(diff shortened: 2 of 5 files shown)


bboure:
Thank you @bsantare for your proposal. 

I would like to get some more background on this issue. What is your use case?

Usually, with serverless (and Cfn, in general), it is considered bad practice to manually change (let alone delete) anything from aws console. This breaks the stack and subsequent deployments might fail.
Any change should only be done through serverless. You proposal looks like a hack to go around that.
I understand we all make mistakes, but I also don't think we should implement mechanisms based on that.
Instead, there should be [recovery procedures](https://aws.amazon.com/premiumsupport/knowledge-center/failing-stack-updates-deleted/) or, alternatively, external tools to "fix" broken stacks.
In your case, for example, you might have tried to remove the ApiKey resource from the template (by choosing another auth method, like Codgnito, or IAM), deploy, then switch back to `API_KEY` and re-deploy again.

Note that this is not limited to Api keys. If you would delete a mapping template, or a data source, the same would happen. I am not sure we want to start implementing this kind of hack for all involved resources.

That being said, I am a bit more concerned about the "expired api key" issue.
AFAIK, expired Api keys remain in the stack (and in AWS console) and should be updated after the following deploy when they *should* be "re-enabled" (extended), as long as the api key was not deleted manually before.
I am not 100% sure about that, though, and I will investigate.
I just created an API with an API key that expires tomorrow (1 day is the minimum lifetime for an API key). Tomorrow, after it expires, I'll try to re-deploy and see if I can reproduce your issue.

If deployment fails, this is definitely an issue (an API key could easily expire without one noticing) and it might be worth exploring solutions.

bsantare:
Thanks for your feedback @bboure.  

This was intended as a recovery procedure in the case this situation presents itself - i.e. set the flag and deploy in order to correct and then remove the flag and re-deploy to get back to steady-state.  AFAIK there is no other way to recover (at present) without completely tearing down the stack.

Deleting an API key/creating a new one (console or CLI) may be used as a means to rotate an API key when necessary.  This would eventually manifest with the same symptom on the next deployment.

bboure:
Hi again,

As expected, re-deploying the API key did work, and the key was re-enabled 

After further investigation, [CloudFormation](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-appsync-apikey.html#cfn-appsync-apikey-apikeyid) supports passing an API key id. 

I am still trying to wrap my head around this one. Changing the param forces Cloudformation to update the given API key (ie: da2-12344567......); **even if the Cfn resource associated to the resource id is a different one.**
However, doing so does **not** re-associate the cfn resource to the given key (it stays attached to the original key) (removing the param after a re-deploy would fail again).

So, in your case, you could just pass the new key in the `ApiKeyId` property and it just works (tested). However, starting from that point you would always need to specify the api key i 'ApiKeyId'.
Also, although deployments would work, the Cfn template would be in a weird state where it would still reference the old (deleted) api key resource. (This is what happened when I tested this).

(Note: This is currently not supported by the plugin.)

> Deleting an API key/creating a new one (console or CLI) may be used as a means to rotate an API key when necessary. This would eventually manifest with the same symptom on the next deployment.

There are plans to [allow more than one API key](https://github.com/sid88in/serverless-appsync-plugin/issues/357). We could imagine a better strategy than manually messing with the stack/cfn template and support rotative api keys.

The way I see it is to add support for *named keys*, so you can specifically add/remove them as needed (renaming a key would just delete/recreate the api key).

Example:

```yml
apiKeys:
  - name: john
    expires: 8760 # hours
  - name: jane
    expires: 8760 # hours
```

if you need to rotate an api key, you then could change it like so 
```yml
apiKeys:
  - name: john2
    expires: 8760 # hours
  - name: jane
    expires: 8760 # hours
```

the old api key for 'john' would be deleted, and a new one would be created (john2). Api key for jane would not change and it would be extended to the new expiry date (today + 8760 hours)

bsantare:
Hello again,

>> As expected, re-deploying the API key did work, and the key was re-enabled

The Appsync docs indicate keys are managed in a Dynamo table and given a TTL (60 days after expiration)
`Expired API keys are kept for 60 days after the expiration time. Key expiration time can be updated while the key is not deleted.`.   I suspect the deployment would fail after 60 days.
https://docs.amazonaws.cn/en_us/appsync/latest/APIReference/API_ApiKey.html

>> However, doing so does not re-associate the cfn resource to the given key (it stays attached to the original key) (removing the param after a re-deploy would fail again).

This was the first thing I tried.  The CFN resource is immutable so associating a key by id during deployment will only work if the AWS::AppSync::ApiKey resource does not already exist in your stack.

The named key option above would solve both issues I believe.  It would be nice if the implementation also allowed specification of a key id so that on manual rotation, you could bump the key definition as you describe above but also include an existing key id along with the new key name.  This I believe will cause the new CFN object to be associated with the existing key.

Another option I considered is an 'eject' param where  `authenticationType === API_KEY` and `unmanagedApiKeys === true` would cause the plugin to not manage API keys at all and allow the user to define/manage their own in CFN.  This seemed to not be in the spirit of the plugin managing everything so I went down the path in this PR.

bboure:
> Expired API keys are kept for 60 days after the expiration time. Key expiration time can be updated while the key is not deleted

Thanks for pointing that out!! I missed that. This issue would be solved by the "named key" feature too. Any key that was deleted manually or automatically by AWS (after 60 days) could just be removed/renamed from the serverless.yml file and it would just be removed from the cfn template too (otherwise, it would fail, yes).
I just tried it quickly: a new api key is created and cfn does not complain about the missing api key.

Manually created api keys could be associated to an existing stack using the "ApiKeyId" field. (Yes, that would be included as an option in a future version). 

> This I believe will cause the new CFN object to be associated with the existing key.

Unfortunately, no, I just added an api key in the cfn template with an `ApiKeyId` that I created manually **before** deploying. The result was that a new api key was created for the new cfn resource.
The api key that I passed in `ApiKeyId` was completely ignored.  After deploying again, `ApiKeyId` was taken into account and the passed api key was updated. The cfn resource is the newly created api key. The manually created api key is not attached to cfn in ay way. 🤷‍♂️ 

By `eject` you mean that the plugin would not generate any API key and all should be handled manually in AppSYnc console? That sounds like a good to have. I don't see anything wrong about it at least.
Ideally, users should pass an array of api keys they want to create. If that list is empty, no api key is generated at all.

I will give some more thoughts to that named keys thing and try to come with a solution. This is a sensible area and we need to make sure not to break anything. 

Thanks for your support! This was useful

bsantare:
Just a small clarification:
> By eject you mean that the plugin would not generate an API key and all should be handled manually in AppSYnc console?

What I meant here is yes API keys are handled manually but via a user's CFN config rather than plugin-managed.  If this was an option, it would also handle the case where the user wants to manage keys through the console or CLI as well.  

I agree if the key definition array can be empty your proposal will handle all of these use cases nicely!

PR #374, opened 2020-12-09
feat(apiKeys): api keys manager

bboure:
Replaces/fixes #373 

Add Api key manager.
Features:
- Allow specifying more than one api key
- Allow custom values for `expiry`, `description` and `apikeyId` (See [Cfn doc](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-appsync-apikey.html)).
- Use names to identify keys.
- Allow op-out of Api key management

### DIFF
--- a/README.md
+++ b/README.md
@@ -74,10 +74,10 @@ custom:
     apiKeys:
       - name: john # name of the api key
         description: 'My api key'
-        expires: 30d # api key life time
+        expiresAfter: 30d # api key life time
       - name: jane
         description: "Jane's api key"
-        expires: 1y
+        expiresAt: '2021-03-09T16:00:00+00:00'
     # Array of additional authentication providers
     additionalAuthenticationProviders:
       - authenticationType: API_KEY
@@ -314,7 +314,7 @@ custom:
 
 #### Managing API keys
 
-Since v1.5.0, api keys management is supported. You can pass one or more api keys as an array in the `appSync.apiKeys` property.
+Since v1.5.0, api keys management is supported. You can pass one or more api keys configuration as an array in the `appSync.apiKeys` property.
 
 The keys can either be a string (name of the key with defaults) or an object of the following shape:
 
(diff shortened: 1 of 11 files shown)


bboure:
@bsantare What do you think? Feedbacks are welcome :)

README.md
```diff
@@ -304,6 +312,52 @@ custom:
...
+
+Since v1.5.0, api keys management is supported. You can pass one or more api keys as an array in the `appSync.apiKeys` property.
+
+The keys can either be a string (name of the key with defaults) or an object of the following shape:
+
+|   property   |      default      | description|
+|--------------| ------------------|------------|
+| name         | *auto-generated*  | Name of the key. This is used under the hood to differentiate keys in the deployment process.<br/><br/>Names are used in the Cfn resource name. Please, keep them short and without spaces or special characters to avoid issues. Key names are case sensitive. |
+| description  | *name of the key*   | A short description for that key |
+| expires      | 1y                | Expiration time for the key. <br/>Can be expressed in seconds or in "human" format. eg: `86400`, `30d`, `2w`, `1y` (See [jkroso/parse-duration](https://github.com/jkroso/parse-duration)).<br/>Min: 1d, max: 1y |
```
bsantare, line 325: This description is a little misleading as the implementation will set the key to expire at now() + `expires` on the next deployment.  Would it make sense to support two modes?

* `expiresOn` - absolute expiration expressed by Posix timestamp or ISO 8601 string
* `expiresAfter` - on each deployment, updated to now() + `expiresAfter` 

At a minimum, I would suggest updating this description to explain that the expires date for the key will be bumped by `expires` on each deployment.

Also, it may be prudent to mention that deployments may fail if a key has expired and was removed via Appsync Dynamo TTL with a note mentioning that the key name may be changed to resolve the broken deployment.

src/index.js
```diff
@@ -426,20 +427,79 @@ class ServerlessAppsyncPlugin {
...
+      const keys = this.getApiKeys(config);
+
+      return keys.reduce((acc, key) => {
+        const {
+          name, expires, description, apiKeyId,
+        } = key;
+
+        let duration = parseDuration(expires);
+        if (duration === null) {
+          throw new Error(`Could not parse ${expires} as a valid diration`);
```
bsantare, line 474: duration

bsantare (commented):
Looks great!

Just one suggestion and a typo

README.md
bboure: > Would it make sense to support two modes? 

Yes!! This is something I thought about yesterday too. Wanted to upgrade the PR but I haven't had time yet.

> Also, it may be prudent to mention that deployments may fail if a key has expired and was removed via Appsync Dynamo TTL with a note mentioning that the key name may be changed to resolve the broken deployment.

Sure, Will add that. Thanks

PR #384, opened 2021-01-31
update Readme with direct lambda section

MarcusJones:


README.md
```diff
@@ -295,6 +295,37 @@ custom:
...
+
+With a direct lambda resolver, no VTL mapping template is required for either request or response. This can be an option if you would like to avoid usage of the Apache VTL langauge or require a complex resolver. Direct lambda resolvers can be enabled for all mappingTemplates with;
+
+```yml
+custom:
+  appsync:
+    defaultMappingTemplates:
+      request: false
+      response: false
+```
```
bboure, line 310: @MarcusJones I would avoid documenting this. As it would work, it can be confusing because `defaultMappingTemplates` applies to ALL resolver types (DynamoDB, HTTP, etc), not only Lambda.

I like the idea to document direct lambda as there currently is only a note as a comment in the Example.
For now, I would just explain to set `false` to each individual Lambda function that requires Direct Lambda support, as you did below.

bboure (approved):
THank you @MarcusJones 
I have adjusted a little the README as per my previous comment. 
I'd prefer to avoid confusions.

Thank you!

Issue #268, opened 2019-08-25
Loggroupname to be API Name instead of API-ID


kvr2277:
Is there anyway to specify log group name to be AppSync's api name instead of API ID. I see this code is used to fetch the loggroupname

` LogGroupName: { 'Fn::Join': ['/', ['/aws/appsync/apis', { 'Fn::GetAtt': [logicalIdGraphQLApi, 'ApiId'] }]] },`

is there a way to make this API name instead. In cloudwatch - pointing at  correct appsync loggroup becomes little difficult as we have to know system generated API ID to search logs.

bboure:
Hi @kvr2277  sorry for late response.
There is currently no way to do what you are looking for, but it would be interesting to extend the `logConfig` option to enable that.

akzincsystems:
This seems to be a sensible suggestion. Every other serverless auto log group formation that I have seen so far names the log group after the functionality, e.g. Lambdas

bboure:
@kvr2277 @akzincsystems After looking at this in #385 this is unfortunately not possible as of now.
This is a CloudFormation limitation. If you look at the [AppSync LogConfig](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-appsync-graphqlapi-logconfig.html) section of the doc, you'll see that there is no way to configure the destination Group name. Aws uses the Api's id by default.

Feel free to open an issue/feature request at https://github.com/aws/aws-appsync-community

I will close this issue for now.

sid88in:
Still the case on v2, and still an AppSync-side constraint: AppSync writes to
`/aws/appsync/apis/<ApiId>` and there's no `LogConfig` field to change the destination group
name. The plugin declares a matching log group only to set retention (`logging.retentionInDays`),
not to rename it — a differently-named group wouldn't receive AppSync's logs.

If you need an API-name-friendly target, the route is forwarding logs (subscription filter)
to your own group; that's outside this plugin. Keeping closed as an AppSync limitation —
worth a request at aws-appsync-community if you'd like AWS to make the group name configurable.

PR #385, opened 2021-02-02
Config for issue #268: API name as LogGroupName

sorenandersen:
Extended `config.logConfig` with new property `logGroupName` that will, if configured to the value `API_NAME`, generate a log group name based on the API name + stage, e.g. "/aws/appsync/apis/api-dev".


bboure:
Thank you @sorenandersen for this PR.
As we discussed outside GitHub, this is not possible. There is no current way to config AppSync to log to a custom Log Group. By default, it's the API key ID.
This change only creates a new LogGroup that will not be used by AppSync

PR #388, opened 2021-02-18
feat: Support updating an existing AppSync API instead of creating a new one

anastyn:
At the moment existing AppSync apis are not supported by the plugin. If there is an existing API in AppSync the plugin will create another one with the same name. 

The proposal is to add back the support of an optional apiId parameter to the Serverless config file. If it is provided then do not create a new AppSync API endpoint but update the existing endpoint instead. This will just update the resources of the existing API and leave the log groups, API URL, API id, authentication provider and the rest of the main stuff unmodified (The auto-generated API URL will not be modified which is important if it is used by some clients and there is no Route53/CloudFront distribution configured).

### DIFF
--- a/README.md
+++ b/README.md
@@ -308,6 +308,9 @@ schema and resolvers from the code.
 If the existing API already contains data sources, those data sources will remain untouched unless they have the same
 names as the data sources in the code, in which case they will be replaced with the ones from the code.
 
+> **Note:** You should never set the apiId of an API that was previously deployed with the same serverless stack, otherwise, it would be deleted. That is because the resource would be removed from the stack.
+>
+> Only use the apiId parameter if you know what you are doing.
 
 ### Multiple APIs
 
@@ -338,7 +341,7 @@ custom:
       ...
 ```
 
-> Note: CloudFormation stack outputs and logical IDs will be changed from the defaults to api name prefixed. This allows you to differentiate the APIs on your stack if you want to work with multiple APIs.
+> **Note:** CloudFormation stack outputs and logical IDs will be changed from the defaults to api name prefixed. This allows you to differentiate the APIs on your stack if you want to work with multiple APIs.
 
 ### Pipeline Resolvers
 
--- a/README.md
+++ b/README.md
@@ -265,6 +265,50 @@ custom:
 
 > Be sure to replace all variables that have been commented out, or have an empty value.
 
+### Working with existing APIs
+
+If you already have an API created in AppSync through the UI or from a different CF stack
+and want to manage it via Serverless then the plugin can also support that.
+
+There is an optional *apiId* parameter that you can use to specify the ID of an existing AppSync API:
+```yaml
+custom:
+  appSync:
+    # ...
+    apiId: 1234abcd
+    # ...
+```
+Without *apiId* parameter the plugin will create a different endpoint with the same name alongside the original one.
+
+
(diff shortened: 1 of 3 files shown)


bboure:
Thank you @anastyn for this! This is indeed a very popular request.
I never really thought about that solution. While it should work, there are a few things to take into consideration here:

- The re-used API will **not** become part of the Cfn stack.
- Authentification, Log Group, Xray and Tags configuration will not be possible through the plugin

All these aspects will have to be handled from the AWS Console.

I also wonder what would happen to existing resolvers/data sources in the pre-existing API when you first re-deploy. Did you test that?

If we go down that road, It would be worth writing some solid documentation that explains all these caveats.

Thanks again

anastyn:
@bboure 
Yes, I tested that.

If there are some existing API data sources the deploy will not remove them, unless they have the same names. So if you query the data sources of the updated API it would include both the old ones and the newly created ones.

Furthermore, if you remove the new CF stack later those newly added data sources will be removed according to the deletion policy, but the old ones created before the first deploy with Serverless AppSync plugin will remain untouched.

The AppSync schema, on the other hand, is completely replaced. So are the resolvers, since they are connected to the schema fields. If you query the resolvers with AWS CLI it will return only the new ones, all old resolvers will be gone.

I can document it.

Later, we can also add a new feature to merge the schema and other resources of the existing API, to manage all of them with the same CF stack. If you think it might be useful feature.
I've been working on some experimental code to merge the existing and the new schemas using graphql tools package. And call AWS CLI to retrieve the old schema, data sources and resolvers, transform them from AppSync JSON format to serverless JSON format and deploy them together as single package. This way it not only deploys the new datasources, resolvers, and schema but also redeploys the existing ones and uses the new CF stack to manage everything.

anastyn:
Added additional documentation explaining the appId property.

anastyn:
@bboure Any suggestions for this merge request?

bboure:
@anastyn No, all looks good. I was hoping to get other feedbacks from people that had this issue in the past (I asked for feedback on issues that were previously open). 

I will try to review this thoughtfully today, but everything looks good to me overall.

I would still add one disclaimer in the doc. 

> You should never set the apiId of an API that was previously deployed with the same serverless stack, otherwise, it would be deleted. That is because the resource would be removed from the stack.
>
> Only use the apiId parameter if you know what you are doing.

anastyn:
Sure, updated the doc

bboure:
LGTM!

Thank you!!

tomaszdudek7:
>Later, we can also add a new feature to merge the schema and other resources of the existing API, to manage all of them with the same CF stack. If you think it might be useful feature.
I've been working on some experimental code to merge the existing and the new schemas using graphql tools package. And call AWS CLI to retrieve the old schema, data sources and resolvers, transform them from AppSync JSON format to serverless JSON format and deploy them together as single package. This way it not only deploys the new datasources, resolvers, and schema but also redeploys the existing ones and uses the new CF stack to manage everything.

Hello @anastyn, is there any work going on with the quoted functionalities? Is your experimental code available anywhere?

I feel like this is a crucial yet missing feature from this plugin. Larger projects have to be split into smaller ones to be more manageable, but this plugin still forces us to create one huge `schema.graphql` file and one huge `mappingTemplates` section, completely breaking most good programming practices.

Issue #395, opened 2021-03-03
Using apiId requires authenticationType and schema.graphql to exist


njeirath:
I'm attempting to use the `apiId` to attach resolvers to an existing API but am getting errors that it still requires that `authenticationType` and `schema.graphql` must exist.  I may just be misunderstanding or misusing the new `apiId` but figured I'd ask the question.

### Background
I have a `backend` serverless project that deploys my Appsync schema and a number of datasources/resolvers/functions.  What I'd like to do is setup a new service say `service1` that can deploy datasources/resolvers/functions related to service1 but tie into the existing schema deployed by `backend`.

When I attempted to use `apiId` only (in addition to my datasources/resolvers/functions) I get the error:
```
AppSync Plugin: Error: appSync property `authenticationType` is missing or invalid.
```

I added `authenticationType: API_KEY` but then it complains about not being able to find `schema.graphql`.  I created an empty one to see if that would work but then CloudFormation complained that the schema was too short (requires a minimum of 1 character).  Finally I tried moving just the parts of the schema that `service1` would be deploying into my `schema.graphql` file but deploying that overwrote the schema that was already there rather than merging the schemas.

### Questions
What I'm curious about then is:

1. I was able to get around having `authenticationType` required when using `apiId` by just setting it to `API_KEY` but I did notice that in the CloudFormation template that was generated it included an `AWS::AppSync::ApiKey` resource.  While having that extra API key is not a huge issue for my use case, I assume it'd be better if `authenticationType` was optional when specifying `apiId`.
2. The bigger issue for me was having to specify `schema.graphql` as there doesn't appear to be any way to tell the plugin to simply use the schema that already exists or, ideally, to merge the schema for `service1` in with the existing schema.

Again I may be just misusing/misunderstanding the purpose of the `apiId` configuration so any help would be greatly appreciated.

bboure:
Hi @njeirath 
The `apiId` field is meant to deploy move an existing API to serverless. It will completely replace all resources of the existing API with those defined in the yml file. 
The error about the missing auth is definitely a bug, since the auth method will be retained by the existing API. 
The error about the schema is correct. You still need a schema that will **replace** the existing one completely.

I I understand correctly what you are trying to achieve is that you want to split your API into several stacks?
The goal is probably to have sort of micro-services that live in the same API?

While it might work (if we removed the constraint on the schema existence), I would not feel confident deploying anything to production with it. 
[...]

PR #396, opened 2021-03-10
Make authentication type optional when api id is used

anastyn:
Made authenticationType attribute optional if user provided the apiId attribute, since it is not used anyway.
It will allow users to create cleaner configs if all they want is to update resources in an existing AppSync API.

Also added 3 more ignored attributes to the README.
And added a log highlighting the same information if user provided an API ID.

Moved the check for the appId attribute to addResources method to make it easier to see from the code where it is used and which attributes are ignored. The ignored attributes are all attributes that modify AppSync API settings.

Updated the unit tests to make sure that the ignored resources are actually ignored.
Tested the creation of a new API and the updates to an existing API.

The attributes are ignored (instead of throwing an exception) due to the backwards compatibility with the previous version, where user could provide the apiId and call the delete operation without removing existing attributes. Otherwise I would add an exception if any of them was provided together with apiId.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -301,19 +301,41 @@ class ServerlessAppsyncPlugin {
     const outputs = this.serverless.service.provider.compiledCloudFormationTemplate.Outputs;
 
     config.forEach((apiConfig) => {
+      this.addResource(resources, outputs, apiConfig);
+    });
+  }
+
+  addResource(resources, outputs, apiConfig) {
+    if (apiConfig.apiId) {
+      this.log(`
+          Updating an existing API endpoint: ${apiConfig.apiId}.
+          The following configuration options are ignored:
+            - name
+            - authenticationType
+            - caching
+            - userPoolConfig
+            - openIdConnectConfig
+            - additionalAuthenticationProviders
+            - logConfig
+            - tags
+            - xrayEnabled
+            - apiKeys
+            - wafConfig
+        `);
+    } else {
       Object.assign(resources, this.getGraphQlApiEndpointResource(apiConfig));
       Object.assign(resources, this.getApiKeyResources(apiConfig));
       Object.assign(resources, this.getApiCachingResource(apiConfig));
-      Object.assign(resources, this.getGraphQLSchemaResource(apiConfig));
       Object.assign(resources, this.getCloudWatchLogsRole(apiConfig));
-      Object.assign(resources, this.getDataSourceIamRolesResouces(apiConfig));
-      Object.assign(resources, this.getDataSourceResources(apiConfig));
-      Object.assign(resources, this.getFunctionConfigurationResources(apiConfig));
-      Object.assign(resources, this.getResolverResources(apiConfig));
       Object.assign(resources, this.getWafResources(apiConfig));
-      Object.assign(outputs, this.getGraphQlApiOutputs(apiConfig));
       Object.assign(outputs, this.getApiKeyOutputs(apiConfig));
-    });
+    }
+    Object.assign(resources, this.getGraphQLSchemaResource(apiConfig));
+    Object.assign(resources, this.getDataSourceIamRolesResouces(apiConfig));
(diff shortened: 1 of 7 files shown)


anastyn:
Any comments?

@bboure maybe some suggestions?

The general idea here is that when the appId is used Serverless maintains only data sources, schema, and resolvers. AppSync configuration is maintained separately in the existing API endpoint settings.

bboure:
LGTM @anastyn Thank you again.

Can we address the missing schema issue also mentioned in #395? If the schema is not present, we just don't deploy the `AWS::AppSync::GraphQLSchema` resource. 

The problem is that this is a very uncommon use case. I'd rather keep an error if no schema is present for the sake of saving most people time and headaches if they just forget to set it.

Can we implement the same logic as for authentication?
- if `apiId` is present, ignore the missing schema
- if `apiId` is **not** present. Throw an error and stop the deployment process.

In fact, this last one is missing I think. I don't see any validation about that. I think it just fails because it tries to read a file that doesn't exist. Can we add a proper validation explaining the issue?

Thank you!

anastyn:
I think we could do that.
But it looks like it's more like a feature than the fix implemented in this pull request.
Can we merge this one, and discuss and review the changes to the schema in another pull request?

src/index.js
```diff
@@ -514,7 +532,7 @@ class ServerlessAppsyncPlugin {
           Type: 'AWS::AppSync::ApiCache',
           Properties: {
             ApiCachingBehavior: config.caching.behavior,
-            ApiId: config.apiId || { 'Fn::GetAtt': [logicalIdGraphQLApi, 'ApiId'] },
+            ApiId: { 'Fn::GetAtt': [logicalIdGraphQLApi, 'ApiId'] },
```
bboure, line 535: @anastyn Why did you remove support for Api Keys and Caching?
As far as I can tell, someone might want to re-use an existing API but still deploy new API keys or configure caching.
This is definitely something possible and that was possible before this change.
I think it is useful to keep support and also not making any breaking change.

anastyn:
@bboure I've just ignored all AppSync settings for simplicity. I think it makes sense to still manage the caching via Serverless, I've added it back. Please check.

For ApiKeys it's a bit more complicated. The authentication type is not modified and not needed when the ApiId is provided, and the existing API may have any authentication type, for example Cognito or OpenID, something different than ApiKey authentication type. 

Even if you add authenticationType attribute to Serverless config it will not update It in AppSync when apiId is provided. So I've removed the keys for simplicity.
The existing AppSync API could be created from any CF template or manually so we can't get its authenticationType attribute from the template.

anastyn:
What we can do is to add a condition and allow modifying apiKeys ignoring the provided authentication type since it is not used anyway, as it was before. Just maybe put some info about it in the doc and in the log.

bboure:
@anastyn Thank you. 
Let's keep it simple for now.
On a side note, I recently came across [this article](https://aws.amazon.com/blogs/aws/new-import-existing-resources-into-a-cloudformation-stack/).
It looks like it can solve most of the issues we're trying to solve with the `apiId` attribute.
I tried it and it works pretty well, although I did not try it with an AppSync stack. With all the resources involved (Api, Schema, resolvers, data-sources), it might become more tedious and complex to do but I think it's doable. 
If you have som time, maybe have a try with it and see if it works in your particular use-case.

bboure:
I resolved the conflit and I will merge now.

Thank you!

anastyn:
@bboure Interesting idea, I'll take a look. But it seems that AppSync resources are not supported at this time.
Here is the full list of the supported resources:

https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/resource-import-supported-resources.html

bboure:
Yep, I figured that out later 😞 
Let's keep an eye on it. For now, your PR will do.
Thanks again

PR #399, opened 2021-04-01
feat: Support glob pattern and absolute path for schema

h-kishi:
Support both glob pattern and absolute path for schema.

```
custom:
  appSync:
    schema:
      - path/**/*.graphql
      - /path/to/schema.graphql
```

Since all schemas are passing through `mergeTypes` by my implementation, the snapshot has changed a lot.


bboure (approved):
Awesome!! Thank you @h-kishi This is a long-waited feature!

PR #406, opened 2021-04-20
Use buildASTSchema for validation

### LINKED ISSUE #402: Use buildASTSchema for validation
From what I can see the validation of graphql schema is done by parsing it, which seems to only check the syntax of the schema and not the semantics. This means that a lot of issues are not caught during validation. The worst of those are errors in auth directives. 

For example I forgot `aws` prefix in `@aws_cognito_user_pools` directive and it passed the validation and got deployed without errors. The problem is that it basically does the opposite of what it's supposed to do, i.e. my query was available using standard api key authentication and threw an error when authenticated with cognito. Another issue is that typos in types also pass validation but the error is thrown later during deployment, which is better but still not ideal.

I looked around for a bit and found that [`buildASTSchema`](https://graphql.org/graphql-js/utilities/#buildastschema) properly validates the schema if stubs for AWS scalars and directives are added (I tested with [python port](https://github.com/graphql-python/graphql-core) of graphql-js, but it should work the same way in the js version; also to make it clear, the stubs are added in validation code not the schema that needs to be validated).  This of course means that if this was to be added the stubs have to be maintained and maybe even an option to fall back to previous validation needs to be added (i.e. `validateSyntaxOnly`).

If you don't quite understand what I'm referring to I can create and link an example that demonstrates this. If I made myself clear enough in this issue and you are open to this change, I could also submit a PR for this when I get time to work on it, which should be withing a couple of days.

AleksaC:
Closes #402 

The rationale behind this PR is mostly explained in the issue. But here is another tl;dr:

```garphql
type Test @cognito_user_pool {
  field: Strin
}
```

The example above (has missing `aws_` prefix in auth directive as well as typo in String type) passes the old validation but fails the new one.

Added config option `validatieSyntaxOnly` which falls back to previous validation behavior. This can be used as an escape hatch until the stubs are updated in case appsync adds new types or changes the existing ones.

### DIFF
--- a/README.md
+++ b/README.md
@@ -624,7 +624,7 @@ custom:
 
 ### Stack Outputs & Exports
 `GraphQlApiId`and `GraphQlApiUrl` are exported to allow cross-stack resource reference using `Fn::ImportValue`.
-Output Exports are named with a `${AWS::StackName}-` prefix to the logical IDs. For example, `${AWS::StackName}-GraphQlApiId`. 
+Output Exports are named with a `${AWS::StackName}-` prefix to the logical IDs. For example, `${AWS::StackName}-GraphQlApiId`.
 > Note: CloudFormation stack outputs and logical IDs will be changed from the defaults to api name prefixed. This allows you to differentiate the APIs on your stack if you want to work with multiple APIs.
 # Cli Usage
 
--- a/src/index.js
+++ b/src/index.js
@@ -1,6 +1,7 @@
 const fs = require('fs');
 const path = require('path');
 const parseSchema = require('graphql/language').parse;
+const { buildASTSchema } = require('graphql/utilities');
 const runPlayground = require('./graphql-playground');
 const getConfig = require('./get-config');
 const chalk = require('chalk');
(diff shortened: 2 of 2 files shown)


bboure:
Thank you @AleksaC 
LGTM. Good job!

I wonder if `validateSyntaxOnly` is really necessary and adds value. If new directives are added, it's rather trivial to add them here I guess?

The serverless framework does something similar with [runtime versions](https://github.com/serverless/serverless/blob/master/lib/plugins/aws/provider.js#L480)

New supported versions have to be added manually later.

What do you think?

AleksaC:
Yeah I was on the fence on this as well. In case appsync introduces new directives they should definitely be added, the `validateSyntaxOnly` was meant to be used if someone is already trying to use them before they are added. This might be too niche of a use case to justify adding the option and I can remove it if you think that's the case.

bboure:
Let's remove it for now. If there is demand/a need for it at some point, we can add it back.

thanks

AleksaC:
Removed it now

bboure:
Awesome Thank you! 🎉

PR #408, opened 2021-04-27
Fix schema path globbing on Windows

### LINKED ISSUE #404: Unable to deploy on windows - Schema Creation Status is FAILED
Have been unable to deploy on windows (currently using gitbash or powershell)

The exact same code is deploying fine using osx.

I have verified that this same behavior is reproducible with the example in this repo and a completely new serverless project adding the bare minimum for serverless-appsync-plugin.

```
Serverless Error ----------------------------------------

  An error occurred: GraphQlSchema - Schema Creation Status is FAILED with details: Found 1 problem(s) with the schema:
  The operation type 'Query' is not present when resolving type 'query' [@3:1]..
 
  Get Support --------------------------------------------
     Docs:          docs.serverless.com
     Bugs:          github.com/serverless/serverless/issues
     Issues:        forum.serverless.com

  Your Environment Information ---------------------------
     Operating System:          win32
     Node Version:              14.12.0
     Framework Version:         2.35.0
     Plugin Version:            4.5.3
     SDK Version:               4.2.2
     Components Version:        3.8.3
```

When comparing the working cloudformation stack template from OSX to the failed template from windows they are identical. 

I have tried yarn and npm to install packages. We have been using serverless-appsync-plugin for a while successfully but failing on a new project. I took the node_modules directory off another windows machine on an older project and replaced the one in the new project and it appears to deploy. However, there are so many differences between the two folders that it's taking a long time to track down which is causing the issue.

Attached is a simple project that fails:
[serverless-appsync-windows-fail.zip](https://github.com/sid88in/serverless-appsync-plugin/files/6332385/serverless-appsync-windows-fail.zip)

AleksaC:
Fixes #404 

The issue is caused by `globby` not working with windows paths as patterns. More precisely `fast-glob` which is what `globby` is based on does not support windows paths as patterns and it seems to be intentional, so I don't think it will change anytime soon. `fast-glob` recommends [replacing `\\` with `/`](https://github.com/mrmlnc/fast-glob#how-to-write-patterns-on-windows) so I did that and it worked.

### DIFF
--- a/src/get-config.js
+++ b/src/get-config.js
@@ -61,13 +61,13 @@ const getConfig = (config, provider, servicePath) => {
   const functionConfigurations = config.functionConfigurations || [];
   const mappingTemplates = config.mappingTemplates || [];
 
-  const toAbsolutePath =
-      filePath => (path.isAbsolute(filePath) ? filePath : path.join(servicePath, filePath));
+  const toAbsolutePosixPath =
+      filePath => (path.isAbsolute(filePath) ? filePath : path.join(servicePath, filePath)).replace(/\\/g, '/');
   const readSchemaFile =
       filePath => fs.readFileSync(filePath, { encoding: 'utf8' });
 
   const schema = Array.isArray(config.schema) ? config.schema : [config.schema || 'schema.graphql'];
-  const schemaFiles = [].concat(...schema.map(s => globby.sync(toAbsolutePath(s))));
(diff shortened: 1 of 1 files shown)


bboure (approved):
LGTM! Thank for a new contrib!

Would you mind fixing this in https://github.com/bboure/serverless-appsync-simulator too?

AleksaC:
Seems very similar at first glance, so I should be able to do it

Issue #425, opened 2021-07-30
AWS_LAMBDA Custom Authorizer support


sweepy84:
AWS have finally released custom authorizer support for AppSync (much like API Gateway) using Lambda as reported > https://aws.amazon.com/about-aws/whats-new/2021/07/aws-appsync-supports-custom-authorization-with-aws-lambda-graphsql-apis/

There is now no need to use Pipeline resolvers to implement custom authorisation. The AWS_LAMBDA option can be set as the default authorizer for all types.

Would be great if the plugin supports this now native feature.

bboure:
Thanks, @sweepy84 

This is definitely something I'd like to see in the plugin
I'll try to get something done in the following days.

harinair:
+1

sid88in:
This shipped in v2. `AWS_LAMBDA` is supported both as the default authorizer and as an
additional auth provider, so you can use it as the default for all types as AWS intended.

Config lives under `authentication` (and `additionalAuthentications`):

```yml
appSync:
  authentication:
    type: 'AWS_LAMBDA'
    config:
      authorizerResultTtlInSeconds: 300
      identityValidationExpression: 'some-regex'
      function:
        handler: 'auth.handler'
```

The plugin emits the `LambdaAuthorizerConfig` on the API and adds the
`lambda:InvokeFunction` permission for AppSync automatically. Docs:
`doc/authentication.md` → "Lambda".

Thanks @sweepy84 for flagging it when AWS first released this. Closing as implemented.

Issue #415, opened 2021-05-07
Error: Unknown type "Query"


alanionita:
Just following the example and getting the following error

```bash
AppSync Plugin: Error: Unknown type "Query".
```

The serverless.yml stack seems to be fine

```yaml
    mappingTemplates:
      - dataSource: testGraphqlApiTableDS
        type: Query
        field: getUser
        request: "getUser-req.vtl"
        response: "common-res.vtl"
      - dataSource: testGraphqlApiTableDS
        type: Mutation
        field: saveUser
        request: "saveUser-req.vtl"
        response: "common-res.vtl"
      - dataSource: testGraphqlApiTableDS
        type: Mutation
        field: deleteUser
        request: "deleteUser-req.vtl"
        response: "common-res.vtl"
```

predescu:
I had the same issue after I've upgraded to the latest version. 
I've reverted back to v1.4.0. Probably other versions works too but I had 1.4.0 previously.

bboure:
@alanionita @predescu 
What does your schema look like?

alanionita:
Schema is pretty basic - just using the demo sample

I actually reverted the versions to `1.10.1` and I'm no longer seeing the above error. 

Seems to be limited to `v1.11.3`.

```graphql
type UserType {
    UserId: ID!
    firstName: String
    lastName: String
    likesMe: Boolean
}


type PaginatedNotes {
    users: [UserType!]!
    nextToken: String
}

type Query {
    getUser(UserId: ID!): UserType
}

type Mutation {
    saveUser(firstName: String!, lastName: String!, likesMe: Boolean): UserType
    deleteUser(UserId: ID!): UserType
}

type Schema {
    query: Query
    mutation: Mutation
}
```

willianrigo:
I'm also facing this problem with version `v1.11.3`.
Rolling back to `1.10.1` and it worked fine.

willianrigo:
Downgrading to `1.10.1` worked

tomaszdudek7:
Is this something that is planned to be fixed? @bboure The plugin does not work from `1.11.3` and above.

I really need `1.12.0` with the https://github.com/sid88in/serverless-appsync-plugin/pull/427 but this issue (that probably started with `1.11.3` as @alanionita mentioned) makes the plugin unusable on anything above `1.11.3`.

bboure:
I am using the latest version in various projects and I have never seen that error.

Is there more context on this error? (e.g. stack trace).

[...]

PR #427, opened 2021-08-05
feat: aws lambda authorizer

mshlz:
Last week, AWS released a new way to authorize AppSync APIs, using Lambda functions like in API Gateway.
[Blog Post here 🔗](https://aws.amazon.com/blogs/mobile/appsync-lambda-auth/)

@sweepy84 mentioned this in that issue (#425).

This pull currently covers only the deployment functionality, allow you to setup the Lambda Authorizer in the `serverless.yml` and deploy to the cloud.

I'm not too familiar with the graphql-playground, cause we use another plugin to execute and test locally here, so I glad if someone could help to implement that too. 😃

This is a snippet how to set up the LAMBDA authorizer. You can refer the function logical name (defined in `functions:`) using the `functionName` field, or pass the Arn directly, (or using the intrinsic function `Fn:GetAtt` to get the Arn).
The `lambdaFunctionArn` has priority, if both `lambdaFunctionArn` and `functionName` are defined.
```yml
custom:
  appSync:
    authenticationType: AWS_LAMBDA
    lambdaAuthorizerConfig:
      functionName: lambdaAuthorizer
      # functionAlias: # optional, used with functionName
      # lambdaFunctionArn: # optional if functionName is present
      #   Fn::GetAtt:
      #     - LambdaAuthorizerLambdaFunction
      #     - Arn
      # authorizerResultTtlInSeconds: 300 # optional
      # identityValidationExpression: .+ # optional
#...
functions:
  lambdaAuthorizer:
    name: MyLambdaAuthorizer
    handler: ./src/functions/LambdaAuthorizer.handler
      
```
In your GraphQL schema you should define the new directive `@aws_lambda`. You can use it together with the others auth directives.

```graphql
type Query @aws_api_key @aws_lambda {
  hello_world: String!
  only_lambda: String!
  @aws_lambda
}
```

Some great and useful links:
https://docs.aws.amazon.com/appsync/latest/devguide/security-authz.html#aws-lambda-authorization
https://docs.aws.amazon.com/appsync/latest/APIReference/API_LambdaAuthorizerConfig.html

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -379,16 +379,24 @@ class ServerlessAppsyncPlugin {
   }
 
   getLambdaAuthorizerConfig(provider) {
-    if (!provider.lambdaAuthorizerConfig || !(provider.lambdaAuthorizerConfig.functionArn || provider.lambdaAuthorizerConfig.functionName)) {
+    if (
+      !provider.lambdaAuthorizerConfig ||
+      !(provider.lambdaAuthorizerConfig.functionArn || provider.lambdaAuthorizerConfig.functionName)
+    ) {
       throw new this.serverless.classes.Error('lambdaAuthorizerConfig `functionArn` or `functionName` must be present');
     }
-    const [functionName, functionAlias] = (provider.lambdaAuthorizerConfig.functionName || '').split(':')
 
     const lambdaAuthorizerConfig = {
-      AuthorizerUri: provider.lambdaAuthorizerConfig.functionArn || this.generateLambdaArn(functionName, functionAlias),
       IdentityValidationExpression: provider.lambdaAuthorizerConfig.identityValidationExpression,
-      AuthorizerResultTtlInSeconds: provider.lambdaAuthorizerConfig.authorizerResultTtlInSeconds
+      AuthorizerResultTtlInSeconds: provider.lambdaAuthorizerConfig.authorizerResultTtlInSeconds,
     };
+
+    if (provider.lambdaAuthorizerConfig.functionArn) {
+      lambdaAuthorizerConfig.AuthorizerUri = provider.lambdaAuthorizerConfig.functionArn;
+    } else {
+      const [functionName, functionAlias] = provider.lambdaAuthorizerConfig.functionName.split(':');
+      lambdaAuthorizerConfig.AuthorizerUri = this.generateLambdaArn(functionName, functionAlias);
+    }
 
     return lambdaAuthorizerConfig;
   }
@@ -468,9 +476,9 @@ class ServerlessAppsyncPlugin {
             Action: 'lambda:InvokeFunction',
             FunctionName: this.getLambdaAuthorizerConfig(config).AuthorizerUri,
             Principal: 'appsync.amazonaws.com',
-            SourceArn: { Ref: logicalIdGraphQLApi }
-          }
-        }
+            SourceArn: { Ref: logicalIdGraphQLApi },
+          },
+        },
       }),
       ...(config.logConfig && config.logConfig.level && {
         [`${logicalIdGraphQLApi}LogGroup`]: {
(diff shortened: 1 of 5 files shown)


src/index.js
```diff
@@ -439,6 +469,17 @@ class ServerlessAppsyncPlugin {
...
+        [`${logicalIdGraphQLApi}LambdaAuthorizerPermission`]: {
+          Type: 'AWS::Lambda::Permission',
+          Properties: {
+            Action: 'lambda:InvokeFunction',
+            FunctionName: this.getLambdaAuthorizerConfig(config).AuthorizerUri,
+            Principal: 'appsync.amazonaws.com',
+            SourceArn: { Ref: logicalIdGraphQLApi },
+          },
+        },
+      }),
```
bboure, line 482: 🔥 awesome!

Aren't we missing the same for `additionalAuthenticationProviders` though?

src/index.js
```diff
@@ -377,6 +378,29 @@ class ServerlessAppsyncPlugin {
...
+
+    const lambdaAuthorizerConfig = {
+      IdentityValidationExpression: provider.lambdaAuthorizerConfig.identityValidationExpression,
+      AuthorizerResultTtlInSeconds: provider.lambdaAuthorizerConfig.authorizerResultTtlInSeconds,
+    };
+
+    if (provider.lambdaAuthorizerConfig.functionArn) {
+      lambdaAuthorizerConfig.AuthorizerUri = provider.lambdaAuthorizerConfig.functionArn;
+    } else {
+      const [functionName, functionAlias] = provider.lambdaAuthorizerConfig.functionName.split(':');
```
bboure, line 397: I definitely like the idea of the `name:alias` notation, however, can we stick to `functionAlias` just like in the [resolver config](https://github.com/sid88in/serverless-appsync-plugin/blob/1cb865f9257edb13be86cfc771067a9facd17b73/src/index.js#L120)? For consistency. 

You probably can reuse `getLambdaArn` around here too.

We can add support for the `:` notation later in a separate PR 🙌

src/index.js
```diff
@@ -439,6 +469,17 @@ class ServerlessAppsyncPlugin {
           Tags: !config.tags ? undefined : this.getTagsConfig(config),
         },
       },
+      ...(config.authenticationType === 'AWS_LAMBDA' && {
+        [`${logicalIdGraphQLApi}LambdaAuthorizerPermission`]: {
+          Type: 'AWS::Lambda::Permission',
+          Properties: {
+            Action: 'lambda:InvokeFunction',
+            FunctionName: this.getLambdaAuthorizerConfig(config).AuthorizerUri,
```
bboure, line 477: Here too, you probably can use something like `getLambdaArn(config.lambdaAuthorizerConfig)`

bboure (commented):
Great job! Thank you @mshlz 🔥 

I left a few comments. Please review an d;et me know what you think.

src/index.js
mshlz: this support will be great! 😃

mshlz:
Thanks @bboure for the review!

I didn't realized that I could get the lambda's reference using that helpers (getLambdaArn(...)) 🤦‍♂️🤯

I'm fixing these issues and testing here. I'll push it soon! 😃

harinair (approved):
Can we merge this and make it available for others to use?

I am currently building an Appsync service and is in need of a Lambda Authorizer

bboure:
Sorry for the delay @mshlz @harinair 

This LGTM! 🚀 

Will be released in a minute!

bboure (approved):

tomaszdudek7:
Sorry for bothering you @mshlz, but have you actually checked whether that change actually works?

We just hit the following issue https://github.com/sid88in/serverless-appsync-plugin/issues/415 (`Unknown type "Query"`) which seems to be linked with version `1.11.3`. Your change was released as `1.12.0`, but when we use that version (and any version that was released after that), we just hit that error from #415, making your PR actually unusable until that issue is fixed.

 Is there any sample code/project to follow and see Lambda Authorizer in practice?

PR #432, opened 2021-08-27
Fix substitutions from "falsy" values

ankon:
The substitutions feature is nice and really helpful, but it couldn't substitute values such as `false`, `""`, and `0`.

This PR fixes that.


bboure:
LGTM!

Thanks a lot @ankon

PR #437, opened 2021-10-09
Better converts and preserves comment descriptions

vicary:
So after some months of tests and production use, I am fairly confident that my fork is good enough for a PR.

This PR adds my [`appsync-schema-converter`](https://gitlab.com/vicary/appsync-schema-converter) to do the followings,
1. Converts string literal descriptions into hash-comment descriptions
2. Converts the separator of multiple interfaces into comma
3. Removes enum descriptions
4. Keep input field descriptions

The current codebase in `serverless-appsync-plugin` removes all comments from the AppSync schema, disabling the self-documenting feat of GraphQL as a whole. One may argue a publicly accessible introspection schema is a security risk, it is really useful as an internal communication tool.

I also added prettier config in `package.json` which matches one of the many conflicting coding style, so there are style changes to make everything consistent. Please let me know if this is not preferred, I can revert those changes back to its original state.

Please comment.

### DIFF
--- a/.prettierrc
+++ b/.prettierrc
@@ -1,4 +1,5 @@
 {
+  "printWidth": 120,
   "tabWidth": 2,
   "useTabs": false,
   "singleQuote": true,
--- a/.prettierrc
+++ b/.prettierrc
@@ -1,8 +1,8 @@
 {
+  "endOfLine": "auto",
   "printWidth": 120,
-  "tabWidth": 2,
-  "useTabs": false,
   "singleQuote": true,
+  "tabWidth": 2,
   "trailingComma": "all",
-  "endOfLine": "auto"
+  "useTabs": false
 }
--- a/.prettierrc
+++ b/.prettierrc
@@ -1,6 +1,6 @@
 {
   "endOfLine": "auto",
-  "printWidth": 120,
+  "printWidth": 80,
   "singleQuote": true,
   "tabWidth": 2,
   "trailingComma": "all",
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,6 @@
 .DS_Store
 node_modules
+node_modules.nosync
 npm-debug.log
 yarn-error.json
 users.json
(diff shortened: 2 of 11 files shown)


vicary:
@bboure alright, it's kinda messy but it's done.

Conflicting lock files got re-installed lazily with no shame, expect minor upgrades but I made sure it's all greens in the test.

A lot of lint fixes snuck in the history because of the new prettier config, let's pretend it's like that from the beginning 😄

I merged my `prettier` section in `package.json` into your `.prettierrc`. Please forgive my OCD for sorting the keys there.

.prettierrc
```diff
@@ -1,7 +1,8 @@
 {
-  "tabWidth": 2,
-  "useTabs": false,
+  "endOfLine": "auto",
+  "printWidth": 120,
```
bboure, line 3: Oh, why 120?

bboure (commented):
Thank you @vicary and sorry for the late review. 
I was wondering why change the prettier config, especially to 120? Any reason?
Most projects use the default (80) or 100.

Side note (that does not have to have any consequences on the final decision):
I personally use panes (like most of us I think) and the max chars I see in one pane on my setup is around 105. So I would have to start scrolling 🙁

vicary:
@bboure Sorry for not aware of the common widths of other IDEs, I have been using Sublime Text and VS Code and have been using `[80, 120]` for my whole life.

120 is picked purely because it makes less of a change compared to the 80 one.

If you like 80 more I don't mind re-formatting it one more time, it's your call.

bboure:
If I'm not mistaken, 80 was the default after I merged #438 (prior to that, it was a bit of a mess, indeed 😄 )

Try changing it back to 80 and you should see less changes in the PR. (It would also make it easier to understand what changed)

vicary:
I must have overlooked, let me change that back in the morning.

vicary:
@bboure It's done, `printWidth` is now 80.

bboure:
LGTM! 🎉 

Thanks again!

PR #440, opened 2021-10-20
fix lint errors

lulzneko:
Fixed a Lint error of "Missing trailing comma".

PR #441, opened 2021-10-20
fix test errors

lulzneko:
## Objective
Fix errors in tests.

## Detail
Jest snapshot seems to be out of date, so the snapshot was updated by `yarn test -u`.


lulzneko:
Lint errors in CI will be resolved in the #440 pull request.

PR #451, opened 2021-11-19
Update serverless minimum requirement

ankon:
This was introduced with #400

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,7 +15,7 @@ Be sure to check out all that [AWS AppSync](https://aws.amazon.com/appsync) has 
 # Minimum requirements
 
 * [Node.js v8 or higher](https://nodejs.org)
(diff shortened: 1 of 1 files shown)


bboure:
Thank you @ankon !

Out of curiosity, what broke?
I checked the PR and it only changes Cloudformation compilation.
How is the sls version breaking it?

ankon:
Nothing broke ;)

I was looking into updating our version, and as part of that read through the PRs and noticed this mismatch. 

Given that both 1.30 and 1.83 are really old this shouldn't really change anything for anyone, I guess.

bboure:
@ankon sorry, left this open for too long. 
I'm still trying to understand where the mismatch comes from though.
Could you clarify?

Thanks

ankon:
Seriously, this is purely a "typo-fix" in the docs to match what `package.json` says.

Issue #453, opened 2021-12-06
The Schema validation/merge is broken


bboure:
Since #437, the schema validator and merger is broken.

The following errors started appearing:
 
```
AppSync Plugin: Error: There can be only one type named "Query".

There can be only one type named "Mutation".
There can be only one type named "Query".
```

The reason is that multiple files define `Query` and `Mutation`

Example:

```yml
appSync:
  schema: *.graphql
```

```graphql
// users.graphql
type Query {
    getUser(id: ID!): User!
}
```

```graphql
// posts.graphql
type Query {
    getPosts(id: ID!): Post!
}
```

**Expected result:**

Schema gets converted to
```
type Query {
    getUser(id: ID!): User!
    getPosts(id: ID!): Post!
}
```

**What happens instead:**

The above error.

@vicary

bboure:
A second possible problem I found with `appsync-schema-converter`:

Literal comments do get transformed into `#` (expected behaviour), but all `#` comments get removed

Example:

```graphql
"""
This comment will remain and be converted to #
"""
type Query {
}

# This comment will disappear
type mutation {
}
```

result

```graphql
# This comment will remain and be converted to #
type Query {
}

type mutation {
}
```

We should probably support "old style" comments

vicary:
@bboure I will see if the new grapgql.js supports merging types the old way, without the extends keyword.

Keeping the comments should be quick, I'll do some tests for both in the weekend.

vicary:
Let me explain a bit more on the schema merging behavior, and how I am planning to solve this.

# The Conflicting Approaches

1. The `graphql-tools` way
It uses [schema merging](https://www.graphql-tools.com/docs/schema-merging#merging-resolvers) via `@graphql-tools/merge#mergeTypeDefs` which expects multiple instances of top level queries from separate files, for example

```graphql
# User.graphql

type User {
  id: ID!
  username: String!
}

type Query {
  users: [User!]!
}
```

```graphql
# Article.graphql

type Article {
  id: ID!
  author: User!
  content: String!
}

type Query {
  articles: [Article!]!
}
```

This method does NOT support the `extend` keyword below when I started `appsync-schema-converter`, I haven't tested again since then.

2. The `graphql.js` way
It parses multiple concatenated files and extends existing types via the `extend` keyword, for example we simply concatenates the following files before conversion:

```graphql
# root.graphql

type Query
type Mutation
```

```graphql
# User.graphql

type User {
  id: ID!
[...]

PR #463, opened 2022-01-14
feat: support MaxBatchSize

vicary:
It caught me by surprise that [MaxBachSize](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-appsync-resolver.html#cfn-appsync-resolver-maxbatchsize) is not supported.

My project recently reached a point where we are actively exploring ways to squeeze out the last bits of performances, [batching](https://docs.aws.amazon.com/appsync/latest/devguide/tutorial-lambda-resolvers.html#advanced-use-case-batching) is definitely something worth trying.

It is also allowed in [Direct Lambda Resolvers](https://docs.aws.amazon.com/appsync/latest/devguide/resolver-mapping-template-reference-lambda.html#direct-lambda-resolvers), scroll down to the section **Direct Lambda Resolvers: Batching enabled** to read more.

### DIFF
--- a/README.md
+++ b/README.md
@@ -163,6 +163,7 @@ custom:
         dataSource: # data source name
         request: # request mapping template name | defaults to {name}.request.vtl
         response: # reponse mapping template name | defaults to {name}.response.vtl
+        maxBatchSize: # maximum number of requests for BatchInvoke operations
     dataSources:
       - type: NONE
         name: none
--- a/README.md
+++ b/README.md
@@ -125,6 +125,7 @@ custom:
           - # function name
         request: # request mapping template name | defaults to `defaultMappingTemplates.request` or {type}.{field}.request.vtl
         response: # response mapping template name | defaults to `defaultMappingTemplates.response` or {type}.{field}.response.vtl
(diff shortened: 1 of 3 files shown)


bboure:
LGTM! Thanks @vicary 

[Pipeline Functions](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-appsync-functionconfiguration.html#cfn-appsync-functionconfiguration-maxbatchsize) also support this.

Could you please add it?

Also, not that we still need to solve [this issue](https://github.com/sid88in/serverless-appsync-plugin/issues/453) before we can release a new version. I will try to find some time to look into this again

vicary:
@bboure Added support for Pipeline Resolvers!

Please take your time on the new major, hopefully a new direction takes shape when you're working on it.

Let me know when you have a new option syntax in your mind, I can help making a new PR for the new schema conversion.

bboure:
@vicary Thanks! LGTM!

I took time to investigate the schema issue yesterday and left feedback on #453

PR #484, opened 2022-03-31
docs: recommend to install as dev dependency

y0ssi10:
As I understand this plugin works at the build stage and may be installed as a dev dependency.


bboure:
Thank you!

PR #498, opened 2022-05-14
docs(offline): fix typo in Readme about offline support libraries

antoinewg:
Was just reading the docs for offline support and found this typo :)


antoinewg:
hi @bboure does this look okey for you ? :)

PR #499, opened 2022-05-21
fix: generate correct dynamodb IamStatement

ebisbe:
When defining 

```
dataSources:
    tableDataSource:
      type: AMAZON_DYNAMODB
      description: 'Main table events for everything. Events, users, guests.'
      config:
        tableName: !Ref Table
```

The generated IamRoleStatement is:

```
"Resource": [
                "arn:aws:dynamodb:eu-west-1:679049160385:table/[Object object]",
                "arn:aws:dynamodb:eu-west-1:679049160385:table/[Object object]/*"
            ],
```

This PR fixes that issue.

### DIFF
--- a/src/__tests__/__snapshots__/dataSources.test.ts.snap
+++ b/src/__tests__/__snapshots__/dataSources.test.ts.snap
@@ -312,18 +312,26 @@ Object {
                 "Resource": Array [
                   Object {
                     "Fn::Join": Array [
-                      ":",
+                      "/",
                       Array [
-                        "arn",
-                        "aws",
-                        "dynamodb",
                         Object {
-                          "Ref": "AWS::Region",
+                          "Fn::Join": Array [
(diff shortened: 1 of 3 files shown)


bboure:
LGTM! Thanks for the first contrib to v2 🎉 

I changed the base branch to `alpha` which is the new main branch for v2. 

Please, would you mind fixing the existing tests, and adding a new one for the use case you fixed?

ebisbe:
When I checked I didn't find related tests to the changes I was doing. If there are files testing that part mind you pointing me the direction? I'll gladly review everything.

Do I have to do anything related with  the change from `v2` to `alpha` ?

ebisbe:
I have added a new test with the particular use case:

bboure:
LGTM 🚀 

Thanks for the contrib!

PR #545, opened 2022-12-23
Implement substitutions for js resolvers

ebisbe:
I've just copy/pasted the MappingTemplate.ts file and did all required refactors. I'm going to test it on my project.

### DIFF
--- a/src/__tests__/js-resolvers.test.ts
+++ b/src/__tests__/js-resolvers.test.ts
@@ -1,5 +1,5 @@
 import { Api } from '../resources/Api';
-import { MappingTemplate } from '../resources/MappingTemplate';
+import { JsResolver } from '../resources/JsResolver';
 import * as given from './given';
 import fs from 'fs';
 
@@ -25,22 +25,23 @@ describe('Mapping Templates', () => {
 
   it('should substitute variables', () => {
     const api = new Api(given.appSyncConfig(), plugin);
-    const mapping = new MappingTemplate(api, {
+    const mapping = new JsResolver(api, {
       path: 'foo.vtl',
       substitutions: {
         foo: 'bar',
         var: { Ref: 'MyReference' },
       },
     });
-    const template =
-      'Foo: ${foo}, Var: ${var}, Context: ${ctx.args.id}, Unknonw: ${unknown}';
+    const template = `const foo = '#foo#';
+       const var = '#var#';
+       const unknonw = '#unknown#'`;
     expect(mapping.processTemplateSubstitutions(template))
       .toMatchInlineSnapshot(`
       Object {
         "Fn::Join": Array [
           "",
           Array [
-            "Foo: ",
+            "const foo = '",
             Object {
               "Fn::Sub": Array [
                 "\${foo}",
@@ -49,7 +50,8 @@ describe('Mapping Templates', () => {
                 },
               ],
             },
-            ", Var: ",
+            "';
+             const var = '",
             Object {
               "Fn::Sub": Array [
                 "\${var}",
@@ -60,7 +62,8 @@ describe('Mapping Templates', () => {
                 },
               ],
             },
-            ", Context: \${ctx.args.id}, Unknonw: \${unknown}",
+            "';
+             const unknonw = '#unknown#'",
           ],
         ],
       }
@@ -77,20 +80,21 @@ describe('Mapping Templates', () => {
       }),
       plugin,
     );
-    const mapping = new MappingTemplate(api, {
+    const mapping = new JsResolver(api, {
       path: 'foo.vtl',
       substitutions: {
         foo: 'fuzz',
       },
     });
-    const template = 'Foo: ${foo}, Var: ${var}';
+    const template = `const foo = '#foo#';
+    const var = '#var#';`;
     expect(mapping.processTemplateSubstitutions(template))
       .toMatchInlineSnapshot(`
       Object {
         "Fn::Join": Array [
           "",
           Array [
-            "Foo: ",
+            "const foo = '",
             Object {
               "Fn::Sub": Array [
                 "\${foo}",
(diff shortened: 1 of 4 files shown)


ebisbe:
You can do a test with your own repository https://github.com/serverless-appsync/appsync-js-resolvers/pull/1

bboure (commented):
Thank you @ebisbe !

I have some suggestions. Let me know what you think.

How about we simply inject _all_ the variables at the top of every file?
The goal would be to avoid trying to detect and replace JS code on the fly, which has several drawbacks:
- prone to errors. (e.g. missing or wrong variable names would generate invalid JS code)
- raw files will always be invalid JS/TS which will give issues with linters and transpilers

The idea is the following:

Users will need to write their resolvers like :

```js
export request() {
  return {
    "operation": "Getitem",
    "key": {
      "pk": MY_VAR
    }
  };
}
```

with substitution:


```yml
resolvers:
  Query.getUser:
    substitutions:
      MY_VAR: !Ref MyRef
```

What we do under the hood is generate the following code:
(using cloudformation `Join` intrinsic function)

```js
const MY_VAR = "MyRefValue";
```

We prepend it to the JS file provided by the user (also using intrinsic function)

This does not solve all the issues though. i.e. linters and transpilers would still complain about inexisting variables before they get added. 

This is a bit tricky, tbh.

ebisbe:
My assumption was that replicating the behaviour that is already tested for the VTL substitutions was fine. It's working. You can see it on the test repository you created. 

> The goal would be to avoid trying to detect and replace JS code on the fly, which has several drawbacks:
>- prone to errors. (e.g. missing or wrong variable names would generate invalid JS code)

The same way that would generate VTL files without the proper values and would fail anyway, right? That's something the user should take care. 
>- raw files will always be invalid JS/TS which will give issues with linters and transpilers

I don't understand that one. The following js resolver is usable and testable. 

```
import { util } from '@aws-appsync/utils';

const test = '#testKey#';
export function request(ctx) {
  return {
    operation: 'GetItem',
    key: {
      name: util.dynamodb.toDynamoDB(ctx.prev.result.name),
    },
  };
}

export function response(ctx) {
  return ctx.result;
}
```
Only point I see with this approach is that they want to use Numeric values. But should be as simple as convert from string to int/float.

ebisbe:
@lengoman I was waiting for a response from @bboure ( I assume he was on holidays ). From my side I think it's the simples implementation we can have right now but we should consider what can be the future of this feature and maybe rework it.

bboure:
Got it thanks @ebisbe 
at one point I was assuming we'd do

```ts
const test = #testKey#;
```
(no quotes) rather than
```ts
const test = '#testKey#';
```

Now I see this is indeed not a problem.

I also think this is the simplest solution for now. 

I was hoping that at some point, AppSync would support env vars for js, but I don't think it will come any time soon. 

What I'm afraid of is that we get into something that will be hard to evolve or change later. 

In fact this discussion made me consider changing the VTL implementation too. 
The goal would be to simplify everything and avoid trying to replace things inline.

e.g.

```vtl
{
  "tableName": $tableName,
}
```
could also become

```vtl
#set($tableName="myTableName")
{
  "tableName": $tableName,
}
```

same for js

```ts
import { util } from '@aws-appsync/utils';

export function request(ctx) {
  return {
    "tableName": MY_TABLE,
  };
}

```
becomes
```ts
import { util } from '@aws-appsync/utils';

const TABLE_NAME = "myTableName";

export function request(ctx) {
  return {
    "tableName": TABLE_NAME,
  };
}

```
(the tricky part here is that is has to go after `import`)

i.e. prepending variable names instead of replacing them so we touch the original code as less as possible.
This would save us all the nested `Fn:Join`s and risks of replacing.

This solution though would be an issue with linters.

For now, I guess your solution works. We can give it a try and see how it goes.
I will try to review this soon and get it merged asap

Thank you again!

ebisbe:
I use VTL and tried Js resolvers and honestly the option to `prepending variable names` feels counter intutitive. 
When testing VTL resolvers I can do: 
```
expect(result).toEqual({
      version: "2018-05-29",
      method: "GET",
      resourcePath: "/services/rest",
      params: {
        query: {
          method: "flickr.tags.getListUserPopular",
          api_key: "${FlickrKey}",
        },
      },
    });
```
and know that I have to declare `FlickrKey` in order for that to be swapped. It's the same principle for the JS resolvers. So I can safely do:
```
expect(result).toEqual({
      version: "2018-05-29",
      method: "GET",
      resourcePath: "/services/rest",
      params: {
        query: {
          method: "flickr.tags.getListUserPopular",
          api_key: "#FlickrKey#",
          user_id: "12345@N67",
          count: 20,
          format: "json",
          nojsoncallback: "1",
        },
      },
    });
    ```
And know that the same declared `FlickrKey` would be swapped here. What I don't like is that we can't use `${}` in the JS resolvers. Well we could but only with single and double quotes which is not ideal. 

We could try to find a better way of replacement that works for both VTL and JS resolvers. Being on alpha lets us find a proper way that it's easier to search/remember.

ebisbe:
@bboure what do you think?

bboure:
Sounds good! sorry @ebisbe  I am quite busy these days. 

What you said also makes sense. 

I definitely want to put more thoughts into this and play trial and error with it.

As you said, being in alpha gives us some flexibility.

bboure:
Found a chance to play with this today. 🎉 
Looks great!  Thanks for the effort!

bboure:
Available in https://github.com/sid88in/serverless-appsync-plugin/releases/tag/v2.0.0-alpha.15 🎉

ebisbe:
Glad it worked and you like it @bboure . If you need  more help with this plugin I will be happy to squeeze some time. I'm using this plugin in a project and I want this to improve.

ebisbe:
@bboure we missed adding the relevant information on the Docs, right? I mean it's not in this PR. I was updating a function to use this and forgot what I did :(

PR #580, opened 2023-03-14
docs: add functions config 

sorenandersen:
Couple of small corrections to the docs.


bboure:
Thank you! Good catch 🎉

Issue #584, opened 2023-03-30
API keys deleted when migrating from v1 -> v2


MatejBalantic:
Our API keys got deleted & recreated when we migrated this plugin from v1 -> v2.

We traced the issue to the naming of the keys in the CloudFormation template.

V1: `GraphQlApiourname`
V2: `GraphQlApiKeyourname`

Note the added `Key` string. This causes CloudFormation to delete existing keys and create new ones.

Luckily we caught this in staging, so we are working around the problem now by manually adding a new key to the AppSync instance via AWS console, using that one until the updated CloudFormation stack is deployed, and then deploying our clients with the newly generated keys. 

This is probably impossible to resolve now, because "fixing" it would break projects using the v2 version, but we should maybe add something to the migration docs?

bboure:
Oh, sorry about that. 
I tried to be backward compatible as much as I could, but this one must have slipped. 

Yes, let's add this to the migration doc. (Would you mind opening a PR?)

Unfortunately, yes, fixing this would become a breaking change in v2, so we can't do that.

A dirty workaround to avoid breaking would be to manually add the Key resource (with the correct logical id) under the serverless.yml `resources`.

sid88in:
thank you!

PR #586, opened 2023-03-31
Add info about API key rotation issue during migration

MatejBalantic:
I've added some info on the API key rotation issue and steps to mitigate it.

Addresses https://github.com/sid88in/serverless-appsync-plugin/issues/584


bboure:
Thanks! 🎉

PR #597, opened 2023-04-26
feat: Support apiId(Multiple cloudformation stacks)

### LINKED ISSUE #595: Single AppSync Api - Multi CloudFormation Stacks - Concurrency issue
Hi, thank you for this plugin!

My team and I are facing a bug when updating the GraphQL schema with multiple stacks while using the plugin. Since we have a huge project, we decided to split it into multiple entities and use the Serverless Compose Framework. Each entity has a dedicated stack with more or fewer resources, which allows us to deploy them individually. Additionally, we have a stack dedicated to AppSync to instantiate the API. The overall structure looks something like this :

```
serverless-compose.yml
services
  |  appsync
  |    |  schema.graphql
  |    |  serverless.ts
  |  entityX
  |    |  getEntityX
  |    |  addEntityX
  |    |  ...
  |    |  serverless.ts
  |  entityY
  |    |  getEntityY
  |    |  addEntityY
  |    |  ...
  |    |  serverless.ts
```
Using the Serverless Compose Framework, we first deploy the AppSync service to provide API information to all other services (Entity X, Entity Y). Then, we deploy all other services simultaneously still using the serverless-appsync-plugin(v1), which is configured to use the API ID. The goal of these services is to add resolvers, data sources, and pipelines to the AppSync API that was deployed in the AppSync service.

However, after deploying all these services (15 for now), we noticed that the GraphQL schema is being updated too frequently, causing some resolvers to become unattached from a mutation/query. When we contacted AWS support, they suggested that too much concurrent updating of the GraphQL schema could be the cause of our problem.

IMHO, the ideal solution would be to have an option to prevent updating the GraphQL schema. This would allow the plugin to only modify external configurations such as resolvers, data sources, and pipelines. It's possible that I have overlooked this solution, and it may have other implications for other features of AppSync. What are your thoughts on this?

(Edit): The goal is to configure the AppSync service to deploy only the schema, while other services do not update it.

HumbleBeck:
Closes #595


src/resources/Api.ts
```diff
@@ -356,26 +392,36 @@ export class Api {
...
-        },
-      };
+    if (
+      !this.config.caching ||
+      this.config.caching?.enabled === false ||
+      this.config.apiId
+    ) {
+      return {};
     }
+    //   if (
```
jeremycare, line 402: You might have forgotten to remove this comment.

Hideman85:
I have the same need, first thank you for your work and this is already helping me.
Now I have the following:
```
// serverless-compose.yml
services:
  service-root:
    path: serverless-definitions/service-root
  service-partA:
    path: serverless-definitions/service-partA
    dependsOn: service-root
    params:
      ParameterApiId: ${service-root.ParameterApiId}
  service-partB:
    path: serverless-definitions/service-partB
    dependsOn: service-root
    params:
      ParameterApiId: ${service-root.ParameterApiId}
```
And I'm getting this error in the other stacks:
```
Template format error: Unresolved resource dependencies [GraphQlSchema, ParameterApiId] in the Resources block of the template
```
Checkking the generated CF and getting on all the resolvers `"DependsOn": ["GraphQlSchema"],` but this resource is only on the root one

Do you think you can have a look at it? Or am I doing something wrong?

HumbleBeck:
hey @Hideman85, indeed, I think I've missed this part thank you Just pushed a commit, it should fix it.

src/types/plugin.ts
```diff
@@ -2,9 +2,9 @@ import { CfnWafRuleStatement, IntrinsicFunction } from './cloudFormation';
 
 export type AppSyncConfig = {
```
simonmeyerrr, line 3: If most of the parameters are ignored when using apiId you should define type like this to avoid misusage of the config:

```
export type BaseAppSyncConfig = {
  dataSources: Record<string, DataSourceConfig>;
  resolvers: Record<string, ResolverConfig>;
  pipelineFunctions: Record<string, PipelineFunctionConfig>;
  substitutions?: Substitutions;
  caching?: CachingConfig;
};

export  type NewAppSyncConfig = BaseAppSyncConfig & {
  name: string;
  schema: string[];
  authentication: Auth;
  additionalAuthentications: Auth[];
  domain?: DomainConfig;
  apiKeys?: Record<string, ApiKeyConfig>;
  xrayEnabled?: boolean;
  logging?: LoggingConfig;
  waf?: WafConfig;
  tags?: Record<string, string>;
};

export type ExistingAppSyncConfig = BaseAppSyncConfig & {
  apiId: string | IntrinsicFunction;
};

export type AppSyncConfig = NewAppSyncConfig | ExistingAppSyncConfig;
```

bboure:
Thanks all for the great feedback and this PR

I will take everything into consideration and have a look at this as soon as I can

Hideman85:
@bboure Sorry to disturb you, have you considered merging this fix for now? Right now I'm using the fork in our flow and this add quite some overhead in our process. This would be awesome to have this released 🙏

morficus:
+1 for getting this merged in. 

my use case is that I have a large AppSync API with v1 of this plugin. we are upgrading to v2 and, at the same time, splitting the API into "services" and using `serverless-compose` to manage them. being able to have an individual "service" update parts of the AppSync API would be phenomenal since it would make things more manageable and speed up deployments. 

thank you @HumbleBeck for the awesome work 🙏

bboure:
Sorry, I'll try to have a look at this over the weekend. 

Meanwhile, AWS released support for [merged APIs](https://aws.amazon.com/blogs/mobile/introducing-merged-apis-on-aws-appsync/). I would like to have a look at it too 

I think both are probably good to support but I'd like to understand how they fit together, etc.

Thanks for the work and your patience

morficus:
@bboure OH WOW 😲 I did not hear about the new "merged API" functionality. AppSync finally has a solution for "schema federation" like Apollo. I'm going to have to dig into the blog post over the weekend. 

Totally agree that it would be great to support both.

morficus:
hey @bboure -- sorry to bug you again 😅 wondering if you were able to put some thought toward this PR or if you had a different approach in mind.

doc/general-config.md
```diff
@@ -185,3 +186,33 @@ appSync:
...
+- domain
+- apiKeys
+- xrayEnabled
+- logging
+- waf
+- Tags
+> Note: you should never specify this parameter if you're managing your AppSync through this plugin since it results in removing your API.
+
+### Schema
+After specifying this parameter, you need to manually keep your schema up to date or from the main stack where your root AppSync API is defined. The plugin is not taking into account schema property due to AppSync limitation and inability to merge schemas across multiple stacks
```
bboure, line 218: This whole section is confusing and probably not necessary. 

Instead, we should add clear guidelines on how and why you should use the `apiId` parameters.
A bit like described [here](https://www.serverless.com/framework/docs/providers/aws/events/apigateway#share-api-gateway-and-api-resources) for API Gateway.

src/index.ts
```diff
@@ -410,6 +421,10 @@ class ServerlessAppsyncPlugin {
   async getIntrospection() {
     const apiId = await this.getApiId();
 
+    if (typeof apiId !== 'string') {
+      return;
```
bboure, line 425: just returning is probably not good enough. 
I would show a warning.

In fact I would probably not allow calling any of these commands (assoc domain, etc) from a "child" service stack.

> Please call this command from the service where the AppSync API is created.

src/index.ts
```diff
@@ -377,6 +384,10 @@ class ServerlessAppsyncPlugin {
   async gatherData() {
     const apiId = await this.getApiId();
 
+    if (typeof apiId !== 'string') {
```
bboure, line 387:
```suggestion
    if (!apiId) {
```

if for some reason, `apiId` is an empty string, it probably does not make sense to continue either.

src/resources/Api.ts
```diff
@@ -40,14 +41,30 @@ export class Api {
...
+            - additionalAuthentications
+            - schema
+            - domain
+            - apiKeys
+            - xrayEnabled
+            - logging
+            - waf
+            - tags
+        `);
+    }
```
bboure, line 59: I'd just validate this.
i.e. don't allow these fields is present in the config along with `apiId`
The validator will take care of showing the warning/error (depending on configuration)

src/resources/Api.ts
```diff
@@ -78,16 +98,17 @@ export class Api {
         Tags: this.getTagsConfig(),
       },
     };
-
-    merge(
-      endpointResource.Properties,
-      this.compileAuthenticationProvider(this.config.authentication),
-    );
+    if (this.config.authentication) {
```
bboure, line 101: `authentication` is always required in this context. 
There is probably some TS magic to do (i.e. a type union type or something)

src/resources/Api.ts
```diff
@@ -525,10 +576,10 @@ export class Api {
   }
 
   hasDataSource(name: string) {
-    return name in this.config.dataSources;
+    return name in (this.config.dataSources || {});
   }
 
   hasPipelineFunction(name: string) {
-    return name in this.config.pipelineFunctions;
+    return name in (this.config.pipelineFunctions || {});
```
bboure, line 583: I understand why you made those optional, but I'd rather keep them as required.
If you look [here](https://github.com/sid88in/serverless-appsync-plugin/blob/05164d8847a554d56bb73590fdc35bf0bda5198e/src/getAppSyncConfig.ts#L36-L46), those are actually already optional from a config point of view.
Then, `getAppSyncConfig()` makes sure to fill them with empty `{}` if needed for when it's injected in the compiler.

src/types/plugin.ts
bboure: I agree on this and it joins what I commented earlier. 
The same should happen in the validation json schema.
I would use something like a union.

```ts
export type BaseAppSyncConfig = {
  dataSources: Record<string, DataSourceConfig>;
  resolvers: Record<string, ResolverConfig>;
  pipelineFunctions: Record<string, PipelineFunctionConfig>;
  substitutions?: Substitutions;
  caching?: CachingConfig;
};

export  type FullAppSyncConfig = BaseAppSyncConfig & {
  name: string;
  schema: string[];
  authentication: Auth;
  additionalAuthentications: Auth[];
  domain?: DomainConfig;
  apiKeys?: Record<string, ApiKeyConfig>;
  xrayEnabled?: boolean;
  logging?: LoggingConfig;
  waf?: WafConfig;
  tags?: Record<string, string>;
};

export  type SharedAppSyncConfig = BaseAppSyncConfig & {
  apiId: string;
};

export type AppSyncConfig = FullAppSyncConfig | SharedAppSyncConfig
```

(not tested, might need adjustments)

src/validation.ts
```diff
@@ -833,8 +833,9 @@ export const appSyncSchema = {
       ],
       errorMessage: 'contains invalid pipeline function definitions',
     },
+    apiId: { $ref: '#/definitions/stringOrIntrinsicFunction' },
   },
-  required: ['name', 'authentication'],
+  required: ['name'],
```
bboure, line 838: as mentioned above, we should have 2 schemas: one for a usual appsync config (main api or single stack) and one for "external API" which only accepts supporters fields.
This way, `authentication` can remain required in single stack scenarios or for main stack and be "forbidden" for sub stacks

bboure:
Thank you for your patience, I had a look at the PR. It looks good but I left some comments for improvements. Espscially on typing and validation. 

A few more thoughts/questions:

- should we allow exporting/importing data sources from other stacks?
e.g. I create my data sources in a main stack (could be a single-table DynamoDB), and then I reference it from resolvers in other stacks.
Right now, we use an [intrinsic function](https://github.com/sid88in/serverless-appsync-plugin/blob/05164d8847a554d56bb73590fdc35bf0bda5198e/src/resources/Resolver.ts#L84) to get the name, which will not work if it's in a different stack. I assume passing the value/name as-is should work. 

The same goes for pipeline functions. 

- We also need to update the [migration guide ](https://github.com/sid88in/serverless-appsync-plugin/blob/master/doc/upgrading-from-v1.md) and remove the reference that using existing APIs is not supported since this is being re-introduced.

- I'd really like to have a full guide on what this is for and when and how to use it. 

Thank you all for the effort!

morficus:
cc: @HumbleBeck ☝️ Check out the comments left on the PR.

morficus:
> I'd really like to have a full guide on what this is for and when and how to use it.

@bboure I can volunteer to help with this part.


> should we allow exporting/importing data sources from other stacks?

my vote on this is no, because as you said, passing the name as-is works. also sharing the same data source between stacks smells like an anti-pattern (unless using single-table design with DynamoDB, which I have not seen people do when using AppSync)

alexandre-snr:
> should we allow exporting/importing data sources from other stacks?

I am interested in this feature, I'd vote yes for the DynamoDB single tables, I feel like this is a very important use case

HumbleBeck:
Hey guys, Thanks for waiting. I was occupied with chores. I'll review the comments by EOW. Meanwhile, can I ask you, @morficus, to help me with documentation/guide? I'll drop this part from my PR.

DaRo0:
I'm really interested in this feature! we can't update our library until this is finished. Just out of curiosity, @HumbleBeck were you able to move forward with your PR? Thanks a lot everyone for the great effort!

BohdanShuliaka:
@HumbleBeck +1 for getting this merged in.

bboure:
What is the status of this @HumbleBeck ?

DaRo0:
@bboure @HumbleBeck did this fell under the table? I don't know if I can help somehow, besides having no knowledge at all

xjnoel:
Hello
I find myself in a case where my AppSync API already exists and this PR meets my need.
What is its progress?

PR #602, opened 2023-06-02
docs: Added Substitution example for JS Resolvers

Lorenzohidalgo:
- Removed duplicated code examples on substitutions.md
- Added Example for JS Resolvers
- Updated Links to substitutions.md to display `Variable Substitutions` instead of `VTL template substitutions`


bboure:
Thank you! 🎉

PR #604, opened 2023-06-13
feat: add visibility configuration

N3RG4L:
Added Visibility property, to add support for Private mode : https://aws.amazon.com/blogs/mobile/introducing-private-apis-on-aws-appsync/

https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-appsync-graphqlapi.html#cfn-appsync-graphqlapi-visibility


bboure (changes requested):
@N3RG4L  Thank you for opening this PR  🙏 🎉

before we merge, could you please
- Add proper TypeScript  types for the new property: (something like `'GLOBAL' | 'PRIVATE'`  instead of just `string`)
- Add [validation](https://github.com/sid88in/serverless-appsync-plugin/blob/master/src/validation.ts)
- Add a few unit tests (generation of Cfn and validation)

N3RG4L:
@bboure  thank you for your fast answer, and for the guidance.
I just added the requirements you mentioned. I hope I missed nothing else

package.json
```diff
@@ -1,9 +1,12 @@
 {
-  "name": "serverless-appsync-plugin",
-  "version": "0.0.0-development",
+  "name": "serverless-appsync-plugin-private",
+  "version": "1.0.0",
```
bboure, line 3: This change looks wrong and will break npm publish

src/validation.ts
```diff
@@ -22,6 +22,11 @@ const DATASOURCE_TYPES = [
   'AMAZON_EVENTBRIDGE',
 ] as const;
 
+const VISIBILITY = [
+  'GLOBAL',
+  'PRIVATE',
+] as const;
```
bboure, line 28: is this used?

src/__tests__/api.test.ts
```diff
@@ -23,6 +23,7 @@ describe('Api', () => {
                 },
               ],
               "XrayEnabled": false,
+              "Visibility": true,
```
bboure, line 26: Shouldn't this be `GLOBAL` or `undefined` ? 
Where does `true` come from?

src/validation.ts
N3RG4L: no, forgot to delete. deleted

src/__tests__/api.test.ts
N3RG4L: I replaced the test with "PRIVATE"

N3RG4L:
added the requested changes

bboure:
I added a few tests and made a few adjustments.
Thank you for your contribution 🎉

bboure (approved):

PR #613, opened 2023-09-16
feat: support js runtime for unit resolver

### LINKED ISSUE #616: Javascript resolvers supported now by AWS as Unit resolvers but not by the plugin
Hey guys,

AWS now supports JS resolvers as UNIT resolvers however if i define them as `kind: UNIT` into the serverless config of your plugin  it throws an error on deploy asking for the Mapping templates(related to VTL resolvers).

If i define them using the `functions:` sintax for pipeline resolvers then it works fine but as pipeline resolvers.

When are you gonna add the support for JS unit resolvers? if it is still supported what's the sintax to write them ? 

Many thanks!

krml4913:
Added support javascript runtime for unit resolver.
https://aws.amazon.com/jp/blogs/mobile/aws-appsync-now-supports-javascript-for-all-resolvers-in-graphql-apis/

Closes #616

### DIFF
--- a/doc/resolvers.md
+++ b/doc/resolvers.md
@@ -33,11 +33,13 @@ appSync:
 - `caching`: [See below](#Caching)
 - `sync`: [See SyncConfig](syncConfig.md)
 
-## JavaScript vs VTL
+## JavaScript, VTL, or Direct Lambda
 
 When `code` is specified, the JavaScript runtime is used. When `request` and/or `response` are specified, the VTL runtime is used.
 
-If neither are specified, by default, the resolver is a PIPELINE JavaScript resolver, and the following minimalistic resolver handler is used.
+For [direct lambda](https://docs.aws.amazon.com/appsync/latest/devguide/direct-lambda-reference.html), set `kind` to `UNIT` and don't specify `request`, `response` or `code`. This only works with Lambda function data sources.
+
+If nothing specified, by default, the resolver is a PIPELINE JavaScript resolver, and the following minimalistic resolver handler is used.
 
 ```js
 export function request() {
@@ -49,7 +51,7 @@ export function response(ctx) {
 }
 ```
 
-If you want to use a UNIT JavaScript resolver, you must specify `code` (the minimalistic resolver handler cannot be used for UNIT resolver).
+Example of a UNIT JavaScript resolver.
 
 ```yaml
 appSync:
@@ -59,8 +61,6 @@ appSync:
       dataSource: myDataSource
       code: getUser.js
 ```
-
-To use [direct lambda](https://docs.aws.amazon.com/appsync/latest/devguide/direct-lambda-reference.html), set `kind` to `UNIT` and don't specify `request`, `response` or `code` (only works with Lambda function data sources).
 
 ## PIPELINE resolvers
 
--- a/doc/resolvers.md
+++ b/doc/resolvers.md
@@ -35,11 +35,13 @@ appSync:
 
 ## JavaScript, VTL, or Direct Lambda
 
-When `code` is specified, the JavaScript runtime is used. When `request` and/or `response` are specified, the VTL runtime is used.
+When `code` is specified, the JavaScript runtime is used.
+
+When `request` and/or `response` are specified, the VTL runtime is used.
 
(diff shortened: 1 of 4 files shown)


src/resources/Resolver.ts
```diff
@@ -31,13 +31,16 @@ export class Resolver {
...
+      this.config.kind === 'UNIT' && 'code' in this.config;
+
+    const isPipelineJsResolver = !(
       this.config.kind === 'UNIT' ||
       'request' in this.config ||
       'response' in this.config
     );
 
-    if (!isJsResolver) {
+    if (!isUnitJsResolver && !isPipelineJsResolver) {
```
bboure, line 43: It's been a while since I wrote this code and with this change it became hard to wrap my head around it.. 

If I read this correctly, we should now be able to simplify this condition to something like `isVTLResolver` where 
```
const isVTLResolver = 'request' in this.config || 'response' in this.config;
```

src/resources/Resolver.ts
```diff
@@ -47,6 +50,16 @@ export class Resolver {
       if (responseMappingTemplate) {
         Properties.ResponseMappingTemplate = responseMappingTemplate;
       }
+    } else {
+      if (this.config.code) {
+        Properties.Code = this.resolveJsCode(this.config.code);
+      } else if (!this.config.code) {
+        Properties.Code = DEFAULT_JS_RESOLVERS;
```
bboure, line 57: This default should only be used for PIPELINE resolvers. 
For UNIT, if neither `request`/`response` nor `code` are present, we should throw an error.

Ideally, we'd want to do that in [validation](https://github.com/sid88in/serverless-appsync-plugin/blob/master/src/validation.ts#L277-L279)
The rule might be a bit complicated.
We probably can do that with an if tule like [here](https://github.com/sid88in/serverless-appsync-plugin/blob/master/src/validation.ts#L80)

bboure (commented):
Thanks @krml4913 !

I left a couple of comments I'd like your feedback on.

src/resources/Resolver.ts
krml4913: Thanks.
I added `isVTLResolver` flag.

src/resources/Resolver.ts
krml4913: `request`, `response`, `code`  [are not necessary for UNIT resolver with direct lambda(datasource)](https://github.com/sid88in/serverless-appsync-plugin/blob/227d479b41ec9b44fc5c1772272d2328395b8ee1/src/__tests__/resolvers.test.ts#L79).
So we cannot add that validation.

Instead, I modified [doc/resolvers.md](https://github.com/sid88in/serverless-appsync-plugin/pull/613/files#diff-38627408e5827a4ac45e70b468a33f7042d8e251c93e774ddfc3769c31ffbd80).

krml4913:
@bboure thank you for review.
I added some commits for your comments and replies.

src/resources/Resolver.ts
bboure: Thanks. 
You're right. I forgot about direct Lambda.

In any case, this still means that if `code` is not present, for `UNIT`, we should NOT use the default. 
Especially if we're using direct Lambda.

doc/resolvers.md
```diff
@@ -49,7 +49,18 @@ export function response(ctx) {
 }
 ```
 
-To use [direct lambda](https://docs.aws.amazon.com/appsync/latest/devguide/direct-lambda-reference.html), set `kind` to `UNIT` and don't specify `request` and `response` (only works with Lambda function data sources).
+If you want to use a UNIT JavaScript resolver, you must specify `code` (the minimalistic resolver handler cannot be used for UNIT resolver).
```
bboure, line 52: What do you mean by "minimalistic" ?

src/resources/Resolver.ts
bboure: I would just remove this condition

`if(isUnitJsResolver || isPipelineJsResolver)`

and add 

`} else if (this.config.kind === 'PIPELINE' && !isVTLResolver) {`

bboure:
Thank you @krml4913

Issue #617, opened 2023-10-03
Upgrading from V1 > V2 drops + creates new API


manwaring:
First of all thank you so much for creating + maintaining this plugin, it's been so helpful!

I went through the steps to upgrade from V1 to V2 and was able to deploy successfully, however during the deployment the prior Appsync API is destroyed and a new one is created.

I found this issue about [v2.3.0 dropping + recreating the API](https://github.com/sid88in/serverless-appsync-plugin/issues/607) but I don't think that's related to what I'm seeing, as I'm not making any changes to my schema or configurations other than those [required for going from V1 to V2](https://github.com/sid88in/serverless-appsync-plugin/blob/master/doc/upgrading-from-v1.md).

Is this a known issue, and/or did I miss a step in the upgrade guide?

bboure:
Hi,

In order to understand what is going on we'd need to check a few things

1. Would you be able to share your v1 and v2 serverless.yml (the relevant parts. i.e. AppSync)
2. Check the generated CloudFormation template from v1 and v2 and look for differences (e.g. different logical id).

manwaring:
It looks like the Logical ID does in fact change from v1 to v2

v1 Logical ID for GraphQLApi
```JSON
  "AppsynctestGraphQlApi": {
    "Type": "AWS::AppSync::GraphQLApi",
    "Properties": {
      "Name": "appsync-test",
      "AuthenticationType": "AMAZON_COGNITO_USER_POOLS",
      "AdditionalAuthenticationProviders": [
        {
          "AuthenticationType": "AWS_IAM"
        }
      ],
      "UserPoolConfig": {
        "AwsRegion": "us-east-1",
        "UserPoolId": {
          "Ref": "UserPool"
        },
        "DefaultAction": "ALLOW"
      },
      "LogConfig": {
        "CloudWatchLogsRoleArn": {
          "Fn::GetAtt": [
            "AppsynctestGraphQlApiCloudWatchLogsRole",
            "Arn"
          ]
        },
        "FieldLogLevel": "ERROR",
        "ExcludeVerboseContent": false
      },
      "XrayEnabled": false
    }
  }
```

v2 Logical ID for GraphQLApi
```JSON
  "GraphQlApi": {
    "Type": "AWS::AppSync::GraphQLApi",
    "Properties": {
      "Name": "appsync-test",
      "XrayEnabled": false,
      "AuthenticationType": "AMAZON_COGNITO_USER_POOLS",
[...]

PR #618, opened 2023-10-03
Add support for specifying a custom Logical ID for the AWS::AppSync::GraphQLApi resource

manwaring:
Exploring a possible solution to #617 by supporting a parameter to prefix the `AWS::AppSync::GraphQLApi` Logical ID (instead of only supporting `GraphQlApi`)

- [x] Update plugin schema/types
- [x] Add optional override property
- [x] Add unit tests
- [ ] Confirm if approach is acceptable to library creator/maintainer

### DIFF
--- a/src/resources/Naming.ts
+++ b/src/resources/Naming.ts
@@ -15,8 +15,8 @@ export class Naming {
     return this.getCfnName(name);
   }
 
-  getApiLogicalId() {
-    return this.getLogicalId(`GraphQlApi`);
+  getApiLogicalId(logicalIdOverride?: string) {
+    return this.getLogicalId(logicalIdOverride || `GraphQlApi`);
   }
 
   getSchemaLogicalId() {
--- a/src/resources/Naming.ts
+++ b/src/resources/Naming.ts
@@ -1,22 +1,24 @@
 import {
+  AppSyncConfig,
   DataSourceConfig,
   PipelineFunctionConfig,
   ResolverConfig,
 } from '../types/plugin';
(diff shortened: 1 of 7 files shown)


src/resources/Naming.ts
```diff
@@ -104,6 +106,6 @@ export class Naming {
   }
 
   getAuthenticationEmbeddedLamdbaName() {
-    return `${this.apiName}Authorizer`;
+    return `${this.config.name}Authorizer`;
```
bboure, line 109: Is this necessary?

Unless I am missing something, we probably could get away with having a 
`apiLogicalId` that overrides `getLogicalId`.

The rest of the logical ids should be safe. i.e. They might be re-created but they are not stateful. 

That would simplify everything.

bboure (commented):
Thank you @manwaring 

Could we also add documentation of when and how to use it?

sid88in:
Closing this in its current form — as noted in review, it makes far more logical IDs
configurable than necessary. Only the `AWS::AppSync::GraphQLApi` logical ID is stateful
(changing it replaces the API); the schema/data-source/resolver/log-group IDs can be
recreated harmlessly, so they don't need to be overridable.

The useful, minimal version of this is a single optional `apiLogicalId` that overrides just
`getApiLogicalId()` — handy for migrating an existing CFN-managed API into the plugin without
forcing a replacement. If anyone wants that, I've opened a scoped `good first issue` for it.

Thanks for the contribution and the prod on the use case!

PR #621, opened 2023-11-02
doc: Fix Resolver example

mapsi:
`kind: UNIT` was missing from the documentation and has cost me quite a bit of time to figure it out, so I should it'd add it. 🤓


bboure:
Thanks!

PR #622, opened 2023-11-09
fix: Exported types

bboure:


ebisbe:
I've test it and it's working again.

PR #623, opened 2023-11-26
fix: schema directives validation

danielfttorres:
As a see below, the `@canonical` directive can be used to annotate that a specific field are canonical and can be reused by an associated API.
Screenshot 2023-11-26 at 16 37 57

But when try to use `@canonical` directive, according to the [AppSync Merged API's spec](https://aws.amazon.com/blogs/mobile/introducing-merged-apis-on-aws-appsync/), results on:
```
Invalid GraphQL schema:
     Unknown directive "@canonical".
```
This PR fix this by adding the support for the `@canonical` directive.

### DIFF
--- a/src/resources/Schema.ts
+++ b/src/resources/Schema.ts
@@ -19,6 +19,7 @@ directive @aws_cognito_user_pools(
   cognito_groups: [String]
 ) on FIELD_DEFINITION | OBJECT
 directive @aws_subscribe(mutations: [String]) on FIELD_DEFINITION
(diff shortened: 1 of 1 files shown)


bboure:
Thank you @danielfttorres 
We should also add the other directives (`@hidden`, `@renamed`, etc).

danielfttorres:
Added the `@hidden` and `@renamed` directives @bboure.

danielfttorres:
@sid88in any new about this?

PR #624, opened 2023-12-13
fix: Avoid Log Role Creation when `roleArn` is proivded

Lorenzohidalgo:
# Detected Issue:
A new Log Role and policy is always created, without considering if one is provided or not. 

If a `roleArn` is provided in the configuration, that one will be used as `CloudWatchLogsRoleArn`.
```js
      set(endpointResource, 'Properties.LogConfig', {
        CloudWatchLogsRoleArn: this.config.logging.roleArn || {
          'Fn::GetAtt': [logicalIdCloudWatchLogsRole, 'Arn'],
        },
        FieldLogLevel: this.config.logging.level,
        ExcludeVerboseContent: this.config.logging.excludeVerboseContent,
      });
```

The current implementation of `compileCloudWatchLogGroup` will always create a new policy and role alongside the log group:
```js
 return {
      [logGroupLogicalId]: {
        Type: 'AWS::Logs::LogGroup',
        ...
      },
      [policyLogicalId]: {
        Type: 'AWS::IAM::Policy',
        ...
      },
      [roleLogicalId]: {
        Type: 'AWS::IAM::Role',
        ...
      },
    };
```

# Proposed Fix:

Update `compileCloudWatchLogGroup` to consider if `roleArn` is provided or not.

We could see two different scenarios:

* `roleArn` is provided -> only the Log Group should be created
* `roleArn` **not** provided -> Log Group, Policy and Role should be created

### DIFF
--- a/src/__tests__/api.test.ts
+++ b/src/__tests__/api.test.ts
@@ -327,6 +327,45 @@ describe('Api', () => {
         }
       `);
     });
+
+    it('should compile CloudWatch Resources without additional IAM Role when logging roleARN is provided', () => {
+      const api = new Api(
+        given.appSyncConfig({
+          logging: {
+            level: 'ERROR',
+            retentionInDays: 14,
+            enabled: true,
+            roleArn: 'arn:',
+          },
+        }),
+        plugin,
+      );
+
+      expect(api.compileCloudWatchLogGroup()).toMatchInlineSnapshot(`
+        Object {
+          "GraphQlApiLogGroup": Object {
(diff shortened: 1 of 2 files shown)


Lorenzohidalgo:
As additional context,

With the current version, deploying with the following configuration
image

Triggers the creation of two roles:
image
Our custom one "AppSyncLoggingServiceRole" and the default one, even though only the custom one will be used.

This fix removes the creation of the default one when a `roleArn` is provided:
image

PR #626, opened 2024-01-19
Support for Amplify directives

lavirez:
If you are using Amplify in the front-end, there comes some directive that are required for code generation and schema generation, I think we need this awsome plugin to pass the those directives to the serverless deployment, so I added those directives to the Schema.ts

### DIFF
--- a/src/resources/Schema.ts
+++ b/src/resources/Schema.ts
@@ -19,6 +19,20 @@ directive @aws_cognito_user_pools(
   cognito_groups: [String]
 ) on FIELD_DEFINITION | OBJECT
 directive @aws_subscribe(mutations: [String]) on FIELD_DEFINITION
+directive @model on OBJECT
+directive @primaryKey on FIELD_DEFINITION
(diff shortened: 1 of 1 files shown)


bboure:
Thanks @lavirez 

Aren't these directives for the Amplify CLI ?
I don't think the Front end (Amplify JS) needs them.
I also think that Amplify CLI removes them before deploying to AppSync, and I'm also pretty sure that AppSync will reject them.

Am I missing something?

lavirez:
> Thanks @lavirez
> 
> Aren't these directives for the Amplify CLI ? I don't think the Front end (Amplify JS) needs them. I also think that Amplify CLI removes them before deploying to AppSync, and I'm also pretty sure that AppSync will reject them.
> 
> Am I missing something?

Thank you @bboure 
You are right these are for Amplify CLI
About the AppSync rejecting the directives, I directly post the Schema from the AppSync console and it works pretty well

The situation is that I need the deployed gpql schema to be downloaded in my Amplify Flutter app.

Issue #632, opened 2024-05-22
Error: Cannot find module 'serverless/lib/serverless-error' after Serverless published version 4.0.23


kumaresdev:
The serverless-appsync-plugin package.json has the following peer dependencies. Serverless has published version 4.0.23, which caused the AppSync deployment to fail.
```
  "peerDependencies": {
    "serverless": ">=3.0.0"
  }
``` 

ERROR:
```
Error: Cannot find module 'serverless/lib/serverless-error'
Require stack:
- /node_modules/serverless-appsync-plugin/lib/resources/Schema.js
- /node_modules/serverless-appsync-plugin/lib/resources/Api.js
- /node_modules/serverless-appsync-plugin/lib/index.js
- /usr/local/lib/node_modules/serverless/lib/utils/require-with-import-fallback.js
- /usr/local/lib/node_modules/serverless/lib/classes/plugin-manager.js
- /usr/local/lib/node_modules/serverless/lib/serverless.js
- /usr/local/lib/node_modules/serverless/scripts/serverless.js
- /usr/local/lib/node_modules/serverless/bin/serverless.js
    at Function.Module._resolveFilename (node:internal/modules/cjs/loader:1028:15)
    at Function.Module._load (node:internal/modules/cjs/loader:873:27)
    at Module.require (node:internal/modules/cjs/loader:1100:19)
    at require (node:internal/modules/cjs/helpers:119:18)
    at Object.<anonymous> (/node_modules/serverless-appsync-plugin/lib/resources/Schema.js:12:42)
    at Module._compile (node:internal/modules/cjs/loader:1198:14)
    at Object.Module._extensions..js (node:internal/modules/cjs/loader:1252:10)
    at Module.load (node:internal/modules/cjs/loader:1076:32)
    at Function.Module._load (node:internal/modules/cjs/loader:911:12)
    at Module.require (node:internal/modules/cjs/loader:1100:19)
    at require (node:internal/modules/cjs/helpers:119:18)
    at Object.<anonymous> (/node_modules/serverless-appsync-plugin/lib/resources/Api.js:31:16)
    at Module._compile (node:internal/modules/cjs/loader:1198:14)
    at Object.Module._extensions..js (node:internal/modules/cjs/loader:1252:10)
    at Module.load (node:internal/modules/cjs/loader:1076:32)
    at Function.Module._load (node:internal/modules/cjs/loader:911:12)
    at Module.require (node:internal/modules/cjs/loader:1100:19)
[...]

PR #635, opened 2024-07-18
fix: support serverless >= 4.0.23 (#632)

timo92:


src/resources/Schema.ts
```diff
@@ -50,7 +49,7 @@ export class Schema {
   valdiateSchema(schema: string) {
     const errors = validateSDL(parse(schema));
     if (errors.length > 0) {
-      throw new ServerlessError(
+      throw new this.api.plugin.serverless.classes.Error(
```
timo92, line 52: Align the way errors are raised with the rest of the repository e.g. https://github.com/sid88in/serverless-appsync-plugin/blob/e99c51ab10a257a051672f53c211bc2a6674bd32/src/resources/Resolver.ts#L83.
The `ServerlessError` import is no longer available starting with serverless 4.0.23.

bboure:
Thank you!

nonso-caleb:
Thanks

PR #636, opened 2024-07-18
fix: Deployment using serverless-compose

timo92:
When using serverless-compose to deploy services the current working directory is set to the cwd of the shell. This also occurs when using the `--config` flag to define the path to the `serverless.yml` in a different directory.

This PR sets the cwd used by globby to find the GraphQL schema files to the directory of the `serverless.yml` of the service which matches the behavior described in the [documentation](https://github.com/sid88in/serverless-appsync-plugin/blob/master/doc/general-config.md#Schema).

### DIFF
--- a/src/resources/Schema.ts
+++ b/src/resources/Schema.ts
@@ -58,13 +58,11 @@ export class Schema {
   }
 
   generateSchema() {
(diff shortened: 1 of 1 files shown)


bboure:
Thank you. This makes sense.

PR #640, opened 2024-09-10
Enable intrinsic function for configs

### LINKED ISSUE #638: [v2 upgrade] Can't use any longer intrinsic functions for config fields like openIdConnectConfig
We're trying to upgrade serverless-appsync-plugin from v1 to v2.

We're using CF functions for some of the values like composing the issuer URL of the OpenID provider:

```ts
additionalAuthenticationProviders: [
  { authenticationType: 'API_KEY' },
  {
    authenticationType: 'OPENID_CONNECT',
    openIdConnectConfig: {
      issuer: {
        'Fn::Join': [
         '',
         [ 'https://', { "Fn::ImportValue": "IssuerDomain" } ]
        ]
      }
    }	
  }
]
```

The serverless-appsync-plugin v1 correctly resovles the issuer URL.

Yet, serverless-appsync-plugin v2 creates below errors:
```
Cannot resolve serverless.ts: Variables resolution errored with:
  - Cannot resolve variable at "resources.Resources.AppSyncApiKeyParameter.Properties.Value": Invalid AppSync Configuration:
     at appSync/additionalAuthentications/1/config/issuer: must be string
  - Cannot resolve variable at "resources.Outputs.GraphQLAppConfigApiKey.Value": Invalid AppSync Configuration:
     at appSync/additionalAuthentications/1/config/issuer: must be string
 ```

We encounter the same issue also with other config fields:
```ts
Cannot resolve serverless.ts: Variables resolution errored with:
  - Cannot resolve variable at "resources.Resources.AppSyncApiKeyParameter.Properties.Value": Invalid AppSync Configuration:
     at appSync/xrayEnabled: must be boolean
     at appSync/waf/enabled: must be boolean
     at appSync/logging/excludeVerboseContent: must be boolean,
```

---

Looking through the validation code, this seems to be an intensional limitation: https://github.com/sid88in/serverless-appsync-plugin/blob/3b56bf248357cd9279062030282dc4cdcde3fcfc/src/validation.ts#L131

Is there a new way to inject computed values into the appsync configuration? Or was this feature removed on purpose?

Happy to contribute PR.

eacet:
This PR enables intrinsic function usage for `xrayEnabled`, `wafEnabled` and `openIdConnectConfig.issuer` fields.

Fixes: https://github.com/sid88in/serverless-appsync-plugin/issues/638

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,10 @@
 {
   "name": "serverless-appsync-plugin",
-  "version": "0.0.0",
   "lockfileVersion": 2,
   "requires": true,
   "packages": {
     "": {
       "name": "serverless-appsync-plugin",
-      "version": "0.0.0",
       "license": "MIT",
       "dependencies": {
         "@graphql-tools/merge": "^8.3.12",
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "serverless-appsync-plugin",
-  "version": "0.0.0",
+  "version": "0.0.1",
   "lockfileVersion": 2,
   "requires": true,
   "packages": {
     "": {
       "name": "serverless-appsync-plugin",
-      "version": "0.0.0",
+      "version": "0.0.1",
       "license": "MIT",
       "dependencies": {
         "@graphql-tools/merge": "^8.3.12",
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "serverless-appsync-plugin",
-  "version": "0.0.1",
+  "version": "0.0.2",
   "lockfileVersion": 2,
   "requires": true,
   "packages": {
     "": {
       "name": "serverless-appsync-plugin",
-      "version": "0.0.1",
+      "version": "0.0.2",
       "license": "MIT",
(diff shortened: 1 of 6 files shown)


sid88in:
Thanks for this, @ronnyroeller — and apologies it sat for so long. 🙏

The codebase has moved on quite a bit since this was opened (the v2 rewrite and the recent AWS SDK v3 migration), so rather than ask you to resolve the conflicts, I've rebased your change onto current master in #713  and credited you as co-author. It keeps your core idea — a booleanOrIntrinsicFunction schema definition alongside the existing stringOrIntrinsicFunction — and applies it to the fields that still exist and pass straight through to CloudFormation: openIdConnectConfig.issuer, xrayEnabled, and logging.excludeVerboseContent.

One adjustment worth noting: wafEnabled from the original no longer exists (it's now waf.enabled), and I deliberately left it — along with logging.enabled, caching.enabled, and domain.enabled — as a plain boolean. Those are synth-time toggles that decide in JS whether the resource is generated at all, so a deploy-time intrinsic can't drive them.
I'll close this in favor of #NNN once that merges. Thanks again for the original fix and for flagging the limitation in #638!

PR #647, opened 2024-11-29
feat: add support for multiple cloudformation stacks (Add apiId)

plezan:
Based on PR #597, updated to be mergeable on master.
I tried to follow @bboure review comments on the original PR.

The goal of this PR is to add support for splited appsync config across multiple services. This is required to bypass the 500 resources limit on cloud formation stacks.

Since the support for Serverless v3 will be dropped at the end of the year, I mainly focused on making those changes compatipble with sls v4.

## Main changes

- Added `apiId` to appsync config to allow adding resources to an existing appsync api
- Handle specific config for existing appsync apis
- Added support for existing data sources
- Use path relative to serverless.yml when using compose

## What else changed in the project ?

To be be properly compatible with serverless v4 and serverless compose i needed to make a few other changes :

- tsconfig now targets es2018
- tsconfig now transpiles to an ESM (previously commonjs)
- Moved validation definition and properties to separate files
- removed a test on a feature that do not exists anymore
- Changed how some errors are handled
- Some methods of the Naming class are now static
- Changed how the api-keys and graphql endpoints are displayed in the console

## Status

Tested with:
- sls v4
- node 20
- vtl templates & substitution
- pipeline functions
- lambda datasources

Not tested with:
- sls v3
- node <20
- JS mapping templates
- custom appsync esbuild settings

The code and documentation still need some cleanup work.

Any feedback is welcome !


plezan:
Also tested without any issue with [serverless-appsync-simulator](https://github.com/serverless-appsync/serverless-appsync-simulator/pull/200) using serverless v4

PR #649, opened 2025-02-05
fix: incorrect path to schema

### LINKED ISSUE #648: graphQLSchema Empty After Upgrading to Serverless Framework v4
After upgrading from Serverless Framework v3.x to v4, the serverless-appsync-plugin is encountering an issue where the graphQLSchema is empty upon deployment. Reverting to v3.x resolves the problem, indicating a potential compatibility issue between the plugin and the latest version of the Serverless Framework.

**Steps to Reproduce:**
- Upgrade Serverless Framework to version 4.x (tested on versions ranging from 4.3.3 to 4.4.18).
- Deploy a service configured with serverless-appsync-plugin and a defined GraphQL schema.
- Observe that the deployed AppSync API has an empty schema, and deployment fails with an error: "Definition cannot be empty."

**Expected Behavior:**
The GraphQL schema should deploy correctly, as it does with Serverless Framework v3.x.

**Actual Behavior:**
The deployed AppSync API has an empty graphQLSchema, and deployment fails with the error mentioned above.

**What I’ve Tried:**

- Tested with multiple Serverless Framework v4 versions, from 4.3.3 to 4.4.18 (latest), with no success.
- Test with multiple serverless-appsync versions from 2.7.0 to 2.9.1, with no success.
- Simplified the GraphQL schema to the following minimal example:
```
type Mutation {
    addTodo(pk: String!, input: TodoInput!): ToDo!
}


type Query {
    getTodo(pk: String!, sk: String!): ToDo
}

type ToDo {
    pk: String!
    sk: String!
    val: String!
}

input TodoInput {
    val: String!
}
```
The issue persists even with this simple schema.
Simplified the configuration to the bare minimum:
```
custom:
  appSync:
    name: 'api'
    schema: 'schema.graphql'
    authentication:
      type: 'AMAZON_COGNITO_USER_POOLS'
      config:
        userPoolId:
          Ref: 'CognitoUserPool'
```
The issue persists with this minimal configuration.
Validated the schema using the validate-schema command, which works fine without errors.
Checked the generated package; the schema definition is missing, indicating it is not being bundled correctly during deployment.

DennisCiba:
After updating to serverless v4 I ran into an issue where the generated schema was always empty.

We are using serverless-compose, so we have a folder structure like this:

```
services/
  service-1/
    serverless.yml
  service-2/
    serverless.yml
serverless-compose.yml
```

It seems that the schema files are paths relative to the service directory, but the `cwd` is the root folder, thus none of the schema files can be found. 

This PR should fix that problem and also resolve #648.

### DIFF
--- a/src/resources/Schema.ts
+++ b/src/resources/Schema.ts
@@ -58,13 +58,16 @@ export class Schema {
   }
 
   generateSchema() {
-    const schemaFiles = flatten(globby.sync(this.schemas));
+    const schemaFiles = flatten(
+      globby.sync(
+        this.schemas.map((schema) =>
+          path.join(this.api.plugin.serverless.config.servicePath, schema),
+        ),
+      ),
+    );
 
     const schemas = schemaFiles.map((file) => {
-      return fs.readFileSync(
-        path.join(this.api.plugin.serverless.config.servicePath, file),
-        'utf8',
-      );
+      return fs.readFileSync(file, 'utf8');
     });
 
     this.valdiateSchema(AWS_TYPES + '\n' + schemas.join('\n'));


PSoltes:
will this be merged at any point?

PR #651, opened 2025-02-21
feat: Support setting evaluationWindowSec for WAF throttling rule

timo92:
AWS has added support for custom evaluation windows for the throttling rule of the Web Application Firewall.

https://aws.amazon.com/about-aws/whats-new/2024/03/aws-waf-rate-based-rules-configurable-time-windows/
https://docs.aws.amazon.com/waf/latest/developerguide/waf-rule-statement-type-rate-based-high-level-settings.html
https://docs.aws.amazon.com/waf/latest/developerguide/waf-rate-based-example-limit-label-aggregation.html

This PR adds support for this feature.

Note: I edited the test snapshots manually to include the property. Let me know if there is a better way

### DIFF
--- a/doc/WAF.md
+++ b/doc/WAF.md
@@ -69,6 +69,7 @@ waf:
     - throttle: 200 # limit to 200 requests per 5 minutes period
     - throttle:
         limit: 200
+        evaluationWindowSec: 300 # count requests in last 5 minutes, one of 60 | 120 | 300 | 600
         priority: 10
         aggregateKeyType: FORWARDED_IP
         forwardedIPConfig:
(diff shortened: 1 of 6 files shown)


sid88in:
need to resolve merge conflict

sid88in:
Thanks for this — clean addition and the AWS mapping is spot-on. One thing to adjust: as written, EvaluationWindowSec is always emitted (defaulting to 300), which adds the property to every existing throttle rule's WebACL on the next deploy. Since 300 is already AWS's default, behavior is unchanged, but AWS resets rate-limiting counts whenever a rate-based setting changes on a live rule, so we'd rather not touch existing stacks. Could we default it to undefined and only emit when set (the way priority/forwardedIPConfig already work)? I've got a patch with that plus a small validation enum (60|120|300|600) so bad values fail at config-validation time instead of at deploy. Also nice catch on the custon→custom typo — that just leaves an obsolete snapshot behind, which a jest -u prunes.

I have made the necessary fixes and gonna merge this in master

sid88in:
fresh PR https://github.com/sid88in/serverless-appsync-plugin/pull/710

Original implementation by @timo92 — rebased onto current master and extended. (Their commit is preserved in history, so authorship is retained.)

PR #663, opened 2025-09-24
when role arn provided it will be used

zjawol:
When roleArn is provided it will be used and new role will not be created.

### DIFF
--- a/src/resources/Api.ts
+++ b/src/resources/Api.ts
@@ -138,7 +138,7 @@ export class Api {
   }
 
   compileCloudWatchLogGroup(): CfnResources {
-    if (!this.config.logging || this.config.logging.enabled === false) {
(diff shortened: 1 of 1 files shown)


AlexHladin:
Hi @zjawol 
Which version of serverless-appsync-plugin are you using? I've checked `roleArn` on version 2.9.1 and it works fine.

src/resources/Api.ts
```diff
@@ -147,64 +147,83 @@ export class Api {
...
-                ],
-                Resource: [
-                  {
-                    'Fn::GetAtt': [logGroupLogicalId, 'Arn'],
-                  },
-                ],
-              },
-            ],
+      };
+    } else {
```
AlexHladin, line 167: @zjawol no need `else` if you have `return` in if statement.

sid88in (approved):

PR #678, opened 2025-10-15
chore: add Node.js 26 to CI

AnastasiyaHladina:
## Proposed changes
This PR adds Node.js 26 version to CI checks. Node.js 26 has been active since [mid-April 2026](https://nodejs.org/en/about/previous-releases), so we should also support it.


AlexHladin:
Node.js 26 is not available for download right now. We should wait for this change until the version is publicly available.

AlexHladin (approved):

sid88in (approved):

PR #704, opened 2026-05-24
test: add CFN synthesis tests covering 24 plugin scenarios

sid88in:
# test: add CFN synthesis tests covering 24 plugin scenarios

## What

Adds a second-tier test suite that exercises the plugin against real
`serverless.yml` configurations and asserts on the generated
CloudFormation, complementing the existing 287 unit tests. Includes 24
runnable example projects under `examples/` that double as user-facing
documentation.

## Why

Unit tests catch logic and validation bugs but can't tell you whether
the plugin actually produces working CloudFormation across the matrix
of features users combine in practice. This adds that layer without
requiring AWS credentials or real deployments — `sls package` does the
synthesis locally and we assert on the template.

Side benefit: the project had no `examples/` folder. Users had to read
the test fixtures or piece together docs to learn how to configure a
feature. These examples are copy-paste-ready and guaranteed to stay
current with the plugin's actual behavior — if they break, CI fails.

## Coverage

24 fixtures, 68 assertions:

**Auth (6):** `basic-api-key`, `cognito-userpools`, `iam-auth`, `oidc-auth`,
`lambda-authorizer`, `multi-auth`

**Resolvers (3):** `lambda-resolvers-js` (JS code with `@aws-appsync/utils`),
`lambda-resolvers-vtl` (VTL templates), `pipeline-resolvers`

**Data sources (6):** `datasource-http` (with both unauthenticated and
IAM-signed variants), `datasource-none`, `datasource-eventbridge`,
`datasource-opensearch`, `datasource-rds`, plus DynamoDB and Lambda
covered transitively through other fixtures

**Advanced features (9):** `caching` (API + per-resolver),
`waf` (rules + association), `logging-xray`, `custom-domain`
(CloudFormation-managed), `introspection-disabled`
(with `queryDepthLimit` and `resolverCountLimit`), `substitutions`,
`environment-variables`, `api-keys-multiple`, `tags`,
`visibility-private`, `schema-multiple-files`

## Infrastructure

- `e2e/helpers/synthesize.ts` — runs `sls package` via a child process,
  returns the parsed CFN template, cleans up temp dirs
- `e2e/helpers/assertions.ts` — readable assertion helpers
  (`expectAuthenticationType`, `expectDataSourceOfType`,
  `expectResourceWithProperties`, etc.)
- `jest.e2e.config.ts` + `jest.e2e.setup.ts` — separate Jest config
  (so unit-test feedback stays fast); `globalSetup` creates the
  `examples/node_modules/serverless-appsync-plugin` symlink so example
  projects resolve the plugin from the current source tree, without
  needing a per-example `npm install`
- `examples/README.md` indexes all 24 examples for users browsing the repo
- `e2e/README.md` explains the testing strategy for contributors

## CI

New `e2e` job in `.github/workflows/ci.yml` runs after the unit-test
matrix on every PR and push to `master` or `alpha`. Single Node version
(22) since CFN output doesn't vary by Node version — runs in ~70 seconds.

## Real plugin behaviors caught while building the fixtures

These are the kind of "silent gotcha" issues the suite is meant to
surface — found and used to refine the examples themselves:

1. `API_KEY` auth silently produces zero API keys if `apiKeys:` array
   isn't specified
2. HTTP data sources with `AWS_IAM` authorization require explicit
   `iamRoleStatements`
3. OpenSearch data sources can't auto-derive the ARN from `Fn::GetAtt`
   — they need an explicit `domain:` field naming the CloudFormation
   resource

## How to run locally

```bash
# Unit tests (unchanged)
npm run test

# CFN synthesis tests
npm run test:e2e

# Everything
npm run test:all

# Single fixture
npx jest --config jest.e2e.config.ts basic-api-key
```

## Verified

- ✅ `npm run test` — 287 unit tests pass
- ✅ `npm run test:e2e` — 68 synthesis assertions across 25 suites pass
- ✅ `npm run lint` — clean (0 errors, 0 warnings)
- ✅ `npm run build` — clean
- ✅ Full suite runs in ~80 seconds locally

## What's not in this PR (intentionally)

- No real-AWS deploy tests (Tier 2) — separate follow-up work that
  needs a dedicated AWS test account and OIDC setup
- No `merged-api` example — the plugin doesn't support `MERGED` API
  type yet; that's a roadmap item, not a missed test

## Follow-ups for after merge

- Real-AWS Tier 2 tests for end-to-end deployment verification on a
  scheduled cadence
- Continue expanding fixtures as new features land (e.g. AppSync Events
  API once supported, MERGED APIs once supported)
- Consider linking from the main README to the `examples/` folder

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -42,3 +42,18 @@ jobs:
             echo "Error: lib directory is empty"
             exit 1
           fi
+
+  e2e:
+    name: CFN Synthesis Tests
+    runs-on: ubuntu-latest
+    needs: tests
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v5
+      - uses: actions/setup-node@v5
+        with:
+          node-version: 22
+      - name: Install dependencies
+        run: npm ci
+      - name: Run CFN synthesis tests
+        run: npm run test:e2e
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -17,11 +17,11 @@ jobs:
       matrix:
         node: [20, 22, 24, 26]
     steps:
-      - uses: actions/setup-node@v5
+      - uses: actions/setup-node@v6
         with:
           node-version: ${{ matrix.node }}
       - name: Checkout code
-        uses: actions/checkout@v5
+        uses: actions/checkout@v6
       - name: Install dependencies
         run: npm ci
       - name: Lint
@@ -49,8 +49,8 @@ jobs:
     needs: tests
     steps:
       - name: Checkout code
-        uses: actions/checkout@v5
-      - uses: actions/setup-node@v5
+        uses: actions/checkout@v6
+      - uses: actions/setup-node@v6
         with:
           node-version: 22
       - name: Install dependencies
--- a/e2e/helpers/assertions.ts
+++ b/e2e/helpers/assertions.ts
@@ -1,4 +1,5 @@
 import { CfnResource, CfnTemplate } from './synthesize';
+import type { AuthenticationType } from '../../src/types/common';
 
 /**
  * Find all resources of a given CloudFormation type.
@@ -96,12 +97,7 @@ function matchesProperties(actual: unknown, expected: unknown): boolean {
  */
 export function expectAuthenticationType(
   template: CfnTemplate,
-  type:
-    | 'API_KEY'
-    | 'AWS_IAM'
-    | 'AMAZON_COGNITO_USER_POOLS'
-    | 'OPENID_CONNECT'
-    | 'AWS_LAMBDA',
+  type: AuthenticationType,
 ): void {
   const { resource } = findOneResourceByType(
     template,
--- a/e2e/helpers/assertions.ts
+++ b/e2e/helpers/assertions.ts
@@ -1,0 +1,159 @@
+import { CfnResource, CfnTemplate } from './synthesize';
+
+/**
+ * Find all resources of a given CloudFormation type.
+ */
+export function findResourcesByType(
+  template: CfnTemplate,
+  type: string,
+): Array<{ logicalId: string; resource: CfnResource }> {
+  return Object.entries(template.Resources)
+    .filter(([, r]) => r.Type === type)
+    .map(([logicalId, resource]) => ({ logicalId, resource }));
+}
+
+/**
+ * Find exactly one resource of a given type and return it. Throws if zero
+ * or more than one match (forces tests to be explicit about cardinality).
+ */
+export function findOneResourceByType(
+  template: CfnTemplate,
+  type: string,
+): { logicalId: string; resource: CfnResource } {
+  const matches = findResourcesByType(template, type);
+  if (matches.length === 0) {
+    throw new Error(`Expected exactly one ${type} resource, found none.`);
+  }
+  if (matches.length > 1) {
+    throw new Error(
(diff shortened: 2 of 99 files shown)


.github/workflows/ci.yml
```diff
@@ -42,3 +42,18 @@ jobs:
...
             exit 1
           fi
+
+  e2e:
+    name: CFN Synthesis Tests
+    runs-on: ubuntu-latest
+    needs: tests
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v5
```
AlexHladin: @sid88in let's switch to the latest v6 version

.github/workflows/ci.yml
```diff
@@ -42,3 +42,18 @@ jobs:
...
+      - uses: actions/setup-node@v5
```
AlexHladin: @sid88in this one, also could be switched to v6

.github/workflows/ci.yml
```diff
@@ -42,3 +42,18 @@ jobs:
...
+        with:
+          node-version: 22
```
AlexHladin, line 55: @sid88in the plugin supports 20, 22, 24, and 26 versions, just curious why 22?

e2e/helpers/assertions.ts
```diff
@@ -0,0 +1,159 @@
...
+ * type configured.
+ */
+export function expectAuthenticationType(
+  template: CfnTemplate,
+  type:
+    | 'API_KEY'
+    | 'AWS_IAM'
+    | 'AMAZON_COGNITO_USER_POOLS'
+    | 'OPENID_CONNECT'
+    | 'AWS_LAMBDA',
```
AlexHladin: @sid88in, how about export this type from [here](https://github.com/sid88in/serverless-appsync-plugin/blob/master/src/validation.ts#L7-L13), and use here?

examples/api-keys-multiple/serverless.yml
```diff
@@ -0,0 +1,36 @@
+service: appsync-api-keys-multiple
+
+provider:
+  name: aws
+  runtime: nodejs20.x
```
AlexHladin: @sid88in nodejs20 is not supported by AWS, only 22 and 24

.github/workflows/ci.yml
sid88in: good catch! bumped

.github/workflows/ci.yml
sid88in: good catch! bumped

.github/workflows/ci.yml
sid88in: CFN output is identical across Node versions. Running 4 Node versions for the synthesis tests would be 4x the CI minutes for zero added signal. The unit-test matrix already verifies the plugin works across Node 20/22/24/26 — those tests exercise the actual plugin logic. The e2e job tests the output of that logic, which doesn't change based on Node.
That said, 22 is somewhat arbitrary — could have been any of the four. For now I will keep it 22

e2e/helpers/assertions.ts
sid88in: good catch fixed

examples/api-keys-multiple/serverless.yml
sid88in: good catch node 20x is EOL April 30, moved to 22

sid88in:
@AlexHladin  please review the CR again. addressed comments - pls verify

AlexHladin (approved):

PR #716, opened 2026-06-01
docs: clarify AppSync custom domain certificate must be in us-east-1

### LINKED ISSUE #587: Automatic certificate generation for custom domains doesn't work in regions other than us-east-1
This plugin creates a certificate in the region of the CloudFormation stack. Our region is `eu-central-1`, which means this is where it gets created.

The problem is that AWS requires certificates for the CloudFront always to be located in `us-east-1`, and in this situation the stack will fail with error:
```
Resource handler returned message: "Invalid request provided: Certificate must be in us-east-1. (Service: AppSync, Status Code: 400)
``

Understandably this plugin can't access regions other than the one in which the CloudFormation stack is. As a result, I don't think that the custom domain certificate generation feature could work in any other region than `us-east-1`.

We've resolved this by generating a certificate manually and referencing it using `domain.certificateArn` configuration.

It would probably make sense to make that clear in the documentation.

### PR DESCRIPTION
**Pull request by sid88in**
## What

Document that the ACM certificate used for an AppSync custom domain **must live in `us-east-1`**, and that the plugin's CloudFormation-managed certificate auto-generation therefore only works when the stack itself is deployed to `us-east-1`.

Closes #587.

## Why

AppSync custom domains are served through a CloudFront distribution, and CloudFront requires its certificate in `us-east-1` — this is an AWS-side constraint that holds regardless of the region the API/stack is deployed to. AWS's own docs are explicit: the certificate "must be located in the ACM us-east-1 AWS Region. No other AWS Region works in this case."

When `useCloudFormation` is `true` and no `certificateArn` is given, the plugin synthesizes an `AWS::CertificateManager::Certificate` resource (`src/resources/Api.ts`, `compileCustomDomain`). CloudFormation only ever creates that resource in the **stack's** region, so for any stack outside `us-east-1` the certificate is created in the wrong region and the deployment fails with:

```
Resource handler returned message: "Invalid request provided: Certificate must be in us-east-1. (Service: AppSync, Status Code: 400)"
```

The reporter (#587) hit this from `eu-central-1` and worked around it by importing a `us-east-1` certificate and referencing it via `domain.certificateArn`. The docs didn't call this requirement out anywhere, so the failure was surprising.

## What this PR does (and does not) do

This is a **documentation-only** change to `doc/custom-domain.md`:

- The `certificateArn` config bullet now states the certificate must be in `us-east-1`.
- The **Certificate** section gains an up-front note that the certificate must be in `us-east-1` (applies to both a provided `certificateArn` and an auto-generated one), and a second note explaining that auto-generation only works for `us-east-1` stacks, with the recommended workaround for other regions.

**Out of scope (intentionally):** auto-creating the certificate *in* `us-east-1` from a stack deployed elsewhere. Plain CloudFormation cannot place a single resource in a different region than its stack; doing so would require either a Lambda-backed custom resource (creating + DNS-validating the cert via the ACM API in `us-east-1`) or a separate `us-east-1` StackSet — a substantial architectural addition, not a surgical change. Documenting the constraint matches what the reporter asked for and keeps the existing single-stack model intact.

## Testing

Docs-only. No source, fixtures, or snapshots are touched, so `build` / `lint` / `test` / `test:e2e` are unaffected (none of them process Markdown). `npx prettier --check doc/custom-domain.md` passes.

## Notes / confidence

- The `us-east-1` requirement is **verified** against AWS documentation and the error in #587.
- I could **not** verify live AWS behavior or Serverless v4 from this environment; the claims here rest on AWS docs and the existing synthesis path in `src/resources/Api.ts`.

## Summary by CodeRabbit

* **Documentation**
  * Expanded custom domain certificate documentation with clarified regional requirements and updated guidance on certificate configuration and deployment constraints.


sid88in:
merging as doc only change, low risk

PR #736, opened 2026-07-02
fix: failing to get correct region when it is configured with variables

gjhommersom:
## Proposed changes

Defer loading of the region until Serverless has resolved the configuration. 
https://www.serverless.com/framework/docs/guides/plugins/creating-plugins#serverless-instance

## Issue(s)

This fixes an issue introduced in version 2.11 that stops the region configuration from being loaded properly. 

## Steps to test or reproduce

Trying to run `serverless deploy` with a configuration like below will fail because the region is invalid.

```
provider:
  name: aws
  region: ${opt:region, param:provider_region}
params:
  default:
    provider_region: eu-west-1

  prod:
    provider_region: eu-central-1
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved plugin startup behavior by delaying cloud configuration lookup until it is actually needed.
  * Reduced the chance of initialization issues when region or credentials are not immediately available.
  * This should make deploy and command execution more reliable in environments with deferred AWS setup.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -120,7 +120,7 @@ type ServerlessPluginUtils = {
 
 class ServerlessAppsyncPlugin {
   private provider: Provider;
-  private clientFactory: AwsClientFactory;
+  private _clientFactory?: AwsClientFactory;
   private gatheredData: {
     apis: {
       id: string;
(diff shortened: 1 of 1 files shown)


sid88in (approved):
lgtm. thanks for the fix!

